PR #8, opened 2023-04-21
Inluce vector in team_tracker.hpp to fix compilation.

malcolmroberts:


BKP:
Hello Malcolm, thanks for the pull request.

We have included the fix into our internal build. The change will go out with the next release (which is scheduled sometime within the next two weeks).

PR #29, opened 2025-01-23
Updated RO builds script and functional test driver

Yiltan:
- Updated RO build scripts
- Updated functional test driver
  - Added a parameter that will accept a hostfile for multi-node jobs
  - Refactored so that we can add `mpirun` flags in a single location
  - Moved signalling APIs tests from short to serial

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -20,577 +20,354 @@
 
 #!/bin/bash
 
-driver_return_status=0
+# This names/values should match the TestType enum in rocSHMEM/tests/functional_tests/tester.hpp
+declare -A TEST_NUMBERS=(
+  ["get"]="0"
+  ["getnbi"]="1"
+  ["put"]="2"
+  ["putnbi"]="3"
+  ["getswarm"]="4"
+  ["amo_fadd"]="5"
+  ["amo_finc"]="6"
+  ["amo_fetch"]="7"
+  ["amo_fcswap"]="8"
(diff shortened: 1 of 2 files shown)


scripts/functional_tests/driver.sh
```diff
@@ -20,577 +20,457 @@
...
+
+DRIVER_RETURN_STATUS=0
+
+ValidateInput $#
+
+case $TEST in
+  ################################################################################
+  ################################ SERIAL TESTS ##################################
+  ################################################################################
+  *"serial")
```
avinashkethineedi: It would be better to organize these tests based on openSHMEM specifications like `RMA`, `Collectives`, `signal_ops` and so on, rather than having everything under `serial`.  Easier to test only certain group of APIs

Yiltan: I tried to follow the previous format to minimize changes as it is already a big PR

That said, I like your idea.  I shall update the PR accordingly. I will also have to update CI to handle the name changes.

avinashkethineedi: The term `serial` has been confusing from the very beginning. I was planning on making those changes in my next PR, but I'll leave it up to you if you'd prefer to include them in this one. 

I do like the current format, it makes the parameters easier to understand.

BKP (dismissed):
It's much more readable.

I agree with Avinash's comment about breaking it into the API subsections from the specification.

avinashkethineedi (approved):
LGTM

PR #30, opened 2025-01-31
Update bandwidth and latency calculations, add multi work group support

avinashkethineedi:
- Refined `bandwidth` and `latency` calculations for improved accuracy
- Added multi work group support for functional tests

### DIFF
--- a/tests/functional_tests/alltoall_tester.cpp
+++ b/tests/functional_tests/alltoall_tester.cpp
@@ -20,8 +20,6 @@
  * IN THE SOFTWARE.
  *****************************************************************************/
 
-using namespace rocshmem;
-
 /* Declare the template with a generic implementation */
 template <typename T>
 __device__ void wg_alltoall(rocshmem_ctx_t ctx, rocshmem_team_t team, T *dest,
@@ -52,30 +50,32 @@ ALLTOALL_DEF_GEN(unsigned int, uint)
 ALLTOALL_DEF_GEN(unsigned long, ulong)
 ALLTOALL_DEF_GEN(unsigned long long, ulonglong)
 
-rocshmem_team_t team_alltoall_world_dup;
-
 /******************************************************************************
  * DEVICE TEST KERNEL
  *****************************************************************************/
 template <typename T1>
-__global__ void AlltoallTest(int loop, int skip, uint64_t *timer,
-                             T1 *source_buf, T1 *dest_buf, int size,
-                             ShmemContextType ctx_type, rocshmem_team_t team) {
+__global__ void AlltoallTest(int loop, int skip, uint64_t *start_time,
+                             uint64_t *end_time, T1 *source_buf, T1 *dest_buf,
+                             int size, ShmemContextType ctx_type,
+                             rocshmem_team_t *teams) {
   __shared__ rocshmem_ctx_t ctx;
+  int wg_id = get_flat_grid_id();
 
   rocshmem_wg_init();
-  rocshmem_wg_ctx_create(ctx_type, &ctx);
+  rocshmem_wg_team_create_ctx(teams[wg_id], ctx_type, &ctx);
 
   int n_pes = rocshmem_ctx_n_pes(ctx);
 
+  source_buf += wg_id * n_pes * size;
+  dest_buf += wg_id * n_pes * size;
+
   __syncthreads();
 
-  uint64_t start;
   for (int i = 0; i < loop + skip; i++) {
     if (i == skip && hipThreadIdx_x == 0) {
-      start = rocshmem_timer();
+      start_time[wg_id] = wall_clock64();
     }
-    wg_alltoall<T1>(ctx, team,
+    wg_alltoall<T1>(ctx, teams[wg_id],
                     dest_buf,    // T* dest
                     source_buf,  // const T* source
                     size);       // int nelement
@@ -84,7 +84,7 @@ __global__ void AlltoallTest(int loop, int skip, uint64_t *timer,
   __syncthreads();
 
   if (hipThreadIdx_x == 0) {
-    timer[hipBlockIdx_x] = rocshmem_timer() - start;
+    end_time[wg_id] = wall_clock64();
   }
 
   rocshmem_wg_ctx_destroy(&ctx);
@@ -95,29 +95,46 @@ __global__ void AlltoallTest(int loop, int skip, uint64_t *timer,
  * HOST TESTER CLASS METHODS
  *****************************************************************************/
 template <typename T1>
-AlltoallTester<T1>::AlltoallTester(
-    TesterArguments args, std::function<void(T1 &, T1 &, T1)> f1,
-    std::function<std::pair<bool, std::string>(const T1 &, T1)> f2)
-    : Tester(args), init_buf{f1}, verify_buf{f2} {
-  int n_pes = rocshmem_team_n_pes(ROCSHMEM_TEAM_WORLD);
-  source_buf = (T1 *)rocshmem_malloc(args.max_msg_size * sizeof(T1) * n_pes);
-  dest_buf = (T1 *)rocshmem_malloc(args.max_msg_size * sizeof(T1) * n_pes);
+AlltoallTester<T1>::AlltoallTester(TesterArguments args)
+    : Tester(args){
+  my_pe = rocshmem_team_my_pe(ROCSHMEM_TEAM_WORLD);
+  n_pes = rocshmem_team_n_pes(ROCSHMEM_TEAM_WORLD);
+
+  int num_elems = (args.max_msg_size / sizeof(T1)) * args.num_wgs * n_pes;
+  int buff_size = num_elems * sizeof(T1);
+
+  source_buf = (T1 *)rocshmem_malloc(buff_size);
+  dest_buf = (T1 *)rocshmem_malloc(buff_size);
+
+  char* value{nullptr};
+  if ((value = getenv("ROCSHMEM_MAX_NUM_TEAMS"))) {
+    num_teams = atoi(value);
+  }
+
+  CHECK_HIP(hipMalloc(&team_alltoall_world_dup,
+                      sizeof(rocshmem_team_t) * num_teams));
 }
 
 template <typename T1>
 AlltoallTester<T1>::~AlltoallTester() {
   rocshmem_free(source_buf);
   rocshmem_free(dest_buf);
+  CHECK_HIP(hipFree(team_alltoall_world_dup));
 }
 
 template <typename T1>
 void AlltoallTester<T1>::preLaunchKernel() {
-  int n_pes = rocshmem_team_n_pes(ROCSHMEM_TEAM_WORLD);
-  bw_factor = sizeof(T1) * n_pes;
+  bw_factor = n_pes;
 
-  team_alltoall_world_dup = ROCSHMEM_TEAM_INVALID;
-  rocshmem_team_split_strided(ROCSHMEM_TEAM_WORLD, 0, 1, n_pes, nullptr, 0,
-                               &team_alltoall_world_dup);
+  for (int team_i = 0; team_i < num_teams; team_i++) {
+    team_alltoall_world_dup[team_i] = ROCSHMEM_TEAM_INVALID;
+    rocshmem_team_split_strided(ROCSHMEM_TEAM_WORLD, 0, 1, n_pes, nullptr, 0,
+                                 &team_alltoall_world_dup[team_i]);
+    if (team_alltoall_world_dup[team_i] == ROCSHMEM_TEAM_INVALID) {
+      std::cout << "Team " << team_i << " is invalid!" << std::endl;
+      abort();
+    }
+  }
 }
 
 template <typename T1>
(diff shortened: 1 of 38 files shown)


tests/functional_tests/alltoall_tester.cpp
```diff
@@ -52,30 +57,33 @@ ALLTOALL_DEF_GEN(unsigned int, uint)
...
+                             uint64_t *end_time, T1 *source_buf, T1 *dest_buf,
+                             int size, ShmemContextType ctx_type,
+                             rocshmem_team_t *teams) {
   __shared__ rocshmem_ctx_t ctx;
+  int wg_id = get_flat_grid_id();
 
   rocshmem_wg_init();
-  rocshmem_wg_ctx_create(ctx_type, &ctx);
+  rocshmem_wg_team_create_ctx(teams[wg_id], ctx_type, &ctx);
+  // rocshmem_wg_ctx_create(ctx_type, &ctx);
```
Yiltan: As discussed, remove comment

tests/functional_tests/alltoall_tester.cpp
```diff
@@ -52,30 +57,33 @@ ALLTOALL_DEF_GEN(unsigned int, uint)
...
+  source_buf += wg_id * n_pes * size;
+  dest_buf += wg_id * n_pes * size;
+
   __syncthreads();
 
-  uint64_t start;
   for (int i = 0; i < loop + skip; i++) {
     if (i == skip && hipThreadIdx_x == 0) {
-      start = rocshmem_timer();
+      start_time[hipBlockIdx_x] = wall_clock64();
```
Yiltan: is `hipBlockIdx_x` the same as `wg_id`?

tests/functional_tests/alltoall_tester.cpp
```diff
@@ -112,52 +123,94 @@ AlltoallTester<T1>::~AlltoallTester() {
...
 void AlltoallTester<T1>::preLaunchKernel() {
-  int n_pes = rocshmem_team_n_pes(ROCSHMEM_TEAM_WORLD);
-  bw_factor = sizeof(T1) * n_pes;
-
-  team_alltoall_world_dup = ROCSHMEM_TEAM_INVALID;
-  rocshmem_team_split_strided(ROCSHMEM_TEAM_WORLD, 0, 1, n_pes, nullptr, 0,
-                               &team_alltoall_world_dup);
+  bw_factor = n_pes;
+
+  for (int team_i = 0; team_i < NUM_TEAMS; team_i++) {
```
Yiltan: 1. I've seen this pattern a few times in this PR, is there a clean way to abstract this? There may not be
2. I was under the impression that #teams == #work group, if so
   2.a. This could be quite slow for RO as we will be calling MPI_Comm_dup 39 times but we may only need 4 teams for a specific functional test (i.e., `mpirun -np 4 ./rocshmem_example_driver -w 4 ...` )
   2.b If these assumptions are true, is it possible to only create the number of teams that the test requires?
3. `rocshmem_team_split_strided` should return a team that is in device memory, is it possible that we create the `team_alltoall_world_dup` array in device memory so that we don't have to issue that memcopy?

tests/functional_tests/alltoall_tester.cpp
```diff
@@ -112,52 +123,94 @@ AlltoallTester<T1>::~AlltoallTester() {
...
-                     _shmem_context, team_alltoall_world_dup);
+                     stream, loop, args.skip, start_time, end_time, source_buf,
+                     dest_buf, num_elems, _shmem_context, teams_on_device);
+
+  num_msgs = (loop + args.skip) * gridSize.x;
+  num_timed_msgs = loop * gridSize.x;
 
-  num_msgs = loop + args.skip;
-  num_timed_msgs = loop;
+  CHECK_HIP(hipFree(teams_on_device));
```
Yiltan: The kernel may not be finished by this moment in time and may still be using this buffer. We could:
1. use `hipFreeAsync` or 
2. Place this free in `postLaunchKernel()` (my preference but it doesn't matter)

tests/functional_tests/amo_extended_tester.cpp
```diff
@@ -40,22 +41,23 @@ __global__ void AMOExtendedTest(int loop, int skip, uint64_t *timer,
...
   rocshmem_free(_r_buf);
   CHECK_HIP(hipFree(_ret_val));
 }
 
 template <typename T>
 void AMOExtendedTester<T>::resetBuffers(uint64_t size) {
-  memset(_r_buf, 0, args.max_msg_size);
-  memset(_ret_val, 0, args.max_msg_size * args.num_wgs);
-  memset(_s_buf, 0, args.max_msg_size * args.num_wgs);
+  memset(_r_buf, 0, sizeof(uint64_t) * args.num_wgs);
```
Yiltan: same as above comment regarding `sizeof(T)` but for the usage of `memset`

tests/functional_tests/amo_extended_tester.cpp
```diff
@@ -95,33 +98,33 @@ void AMOExtendedTester<T>::verifyResults(uint64_t size) {
...
   __global__ void AMOExtendedTest<T>(                                         \
-      int loop, int skip, uint64_t *timer, char *r_buf, T *s_buf, T *ret_val, \
-      TestType type, ShmemContextType ctx_type) {                             \
+      int loop, int skip, uint64_t *start_time, uint64_t *end_time,           \
+      char *r_buf, T *s_buf, T *ret_val, TestType type,                       \
+      ShmemContextType ctx_type) {                                            \
     __shared__ rocshmem_ctx_t ctx;                                            \
     rocshmem_wg_init();                                                       \
     rocshmem_wg_ctx_create(ctx_type, &ctx);                                   \
+    int blk_idx = get_flat_grid_id();                                         \
```
Yiltan: Could be nice to have this as `wg_id` to keep it consistent with other tests. Some comment elsewhere in this PR too

tests/functional_tests/barrier_all_tester.cpp
```diff
@@ -67,9 +68,9 @@ void BarrierAllTester::launchKernel(dim3 gridSize, dim3 blockSize, int loop,
   size_t shared_bytes = 0;
 
   hipLaunchKernelGGL(BarrierAllTest, gridSize, blockSize, shared_bytes, stream,
-                     loop, args.skip, timer);
+                     loop, args.skip, start_time, end_time);
 
-  num_msgs = (loop + args.skip) * gridSize.x;
+  num_msgs = loop + args.skip;
```
Yiltan, line 73: As this discussed in a meeting I understand why we don't multiply by `gridSize` in this test but not sure if the AMO is the same case.

tests/functional_tests/extended_primitives.cpp
```diff
@@ -31,26 +31,25 @@ using namespace rocshmem;
...
   rocshmem_wg_ctx_create(ctx_type, &ctx);
 
   /**
-   * Calculate start index for each work group for tiled version
-   * If the number of work groups is greater than 1, this kernel performs a
-   * tiled functional test
-  */
-  uint64_t start;
+   * Calculate start index for each work group
+   */
```
Yiltan: Comment on a single line

tests/functional_tests/extended_primitives.cpp
```diff
@@ -31,26 +31,25 @@ using namespace rocshmem;
...
   uint64_t idx = size * get_flat_grid_id();
   s_buf += idx;
   r_buf += idx;
 
   for (int i = 0; i < loop + skip; i++) {
-    if (i == skip) start = rocshmem_timer();
+    if (i == skip) {
+        start_time[hipBlockIdx_x] = wall_clock64();
+    }
```
Yiltan, line 52: Sometimes we use 
```
  if (i == skip) {
        start_time[hipBlockIdx_x] = wall_clock64();
  }
```
Sometimes we use 
```
  if (i == skip) start_time[hipBlockIdx_x] = wall_clock64();
```

Either works for me but we could make it consistent

tests/functional_tests/alltoall_tester.cpp
```diff
@@ -112,52 +123,94 @@ AlltoallTester<T1>::~AlltoallTester() {
...
+  for(int wg_id = 0; wg_id < args.num_wgs; wg_id++) {
+    for(int pe = 0; pe < n_pes; pe++) {
+      for(int i = 0; i < num_elems; i++) {
+        idx = (wg_id * n_pes + pe) * num_elems + i;
+        if (dest_buf[idx] != source_buf[idx]) {
+          std::cerr << "Data validation error at idx " << idx << std::endl;
+          std::cerr << "PE " << my_pe << " Got " << dest_buf[idx]
+          << ", Expected " << source_buf[idx] << std::endl;
+          // exit(-1);
+          return;
```
Yiltan: With `return` does it still propagate the failure to the driver script like `exit(-1)` does?

tests/functional_tests/fcollect_tester.cpp
```diff
@@ -112,54 +140,92 @@ FcollectTester<T1>::~FcollectTester() {
...
+        }
+        else if constexpr (std::is_integral<T1>::value) {
+          expected = pe;
+        }
+        if (dest_buf[idx] != expected) {
+          std::cerr << "Data validation error at idx " << idx << std::endl;
+          std::cerr << "PE " << my_pe << " Got " << dest_buf[idx]
+          << ", Expected " << expected << std::endl;
+          // exit(-1);
+          return;
```
Yiltan: Same comment as earlier - With `return` does it still propagate the failure to the driver script like `exit(-1)` does?

tests/functional_tests/extended_primitives.cpp
```diff
@@ -105,8 +104,8 @@ void ExtendedPrimitiveTester::launchKernel(dim3 gridSize, dim3 blockSize,
   size_t shared_bytes = 0;
 
   hipLaunchKernelGGL(ExtendedPrimitiveTest, gridSize, blockSize, shared_bytes,
-                     stream, loop, args.skip, timer, (char*)s_buf,
-                     (char*)r_buf, size, _type, _shmem_context);
+                     stream, loop, args.skip, start_time, end_time,
+                     (char*)s_buf, (char*)r_buf, size, _type, _shmem_context);
```
Yiltan, line 108: In line 88, we could allocate as `char` as we consistently cast to it

tests/functional_tests/tester.cpp
```diff
@@ -352,35 +319,22 @@ std::vector<Tester*> Tester::create(TesterArguments args) {
       testers.push_back(new ShmemPtrTester(args));
       return testers;
     case WGGetTestType:
-      if (rank == 0) {
-        if (args.num_wgs > 1)
-          std::cout << "Tiled Blocking WG level Gets ###" << std::endl;
-        else std::cout << "Blocking WG level Gets ###" << std::endl;
-      }
+      if (rank == 0)
+        std::cout << "Blocking WG level Gets ###" << std::endl;
```
Yiltan, line 323: Thank you for this - I think I broke this when I updated the `driver.sh` script

tests/functional_tests/tester.cpp
```diff
@@ -583,33 +502,43 @@ void Tester::print(uint64_t size) {
...
+      static_cast<double>(total_size * bw_factor) / time_s / pow(2, 30);
 
   int field_width = 20;
   int float_precision = 2;
 
   if (_print_header) {
-    printf("%-*s%*s%*s%*s",
-           10, "# Size (B)",
+    printf("%-*s%-*s%*s%*s%*s",
+           15, "# Size (B)",
```
Yiltan, line 527: Thoughts on adding `Total Size (B)` which would be `total_size` as this would reflect the volume of data transferred? Or maybe we adjust the `Size (B)`, I'm not sure what's best

Yiltan (commented):
Looks good, mostly very minor comment or question about my understanding

tests/functional_tests/tester.cpp
avinashkethineedi: The number of messages varies depending on the API being tested, and I also wanted to keep the size to indicate the amount of data transferred per message. The total size would be a multiple of these two. We could include the total size as well, but while it might be reasonable for RMA operations, I'm unsure if it makes sense to keep it for collectives and atomics.

tests/functional_tests/barrier_all_tester.cpp
avinashkethineedi: It is necessary for AMO since each thread block operates on a different variable.

tests/functional_tests/tester.cpp
Yiltan: Ah didn't think about atomics - yeah, leave this as is then.

tests/functional_tests/extended_primitives.cpp
avinashkethineedi: The `r_buf` and `s_buf` buffers are declared as `int` because they store contiguous integer values used for result verification after the rocSHMEM RMA APIs transfer data to `r_buf`. However, within the kernels, the `getmem/putmem` APIs handle data movement and require the pointer to be cast to `char*`. This is necessary because these APIs interpret the size parameter as the number of bytes to transfer; without casting, the pointer would be treated as an `int*`, leading to unintended integer-based pointer arithmetic.

tests/functional_tests/alltoall_tester.cpp
avinashkethineedi: Moved teams creation and deletion to the `preLaunchKernel` and `postLaunchKernel` functions, respectively.

tests/functional_tests/alltoall_tester.cpp
avinashkethineedi: - For collective operations, the number of teams should match the number of work groups participating in the operation. This may slow down the `init` phase for the RO backend, as it requires creating multiple MPI windows.  
- Updated the logic so that the number of teams created will now be determined by the `ROCSHMEM_MAX_NUM_TEAMS` environment variable if set; otherwise, it defaults to `39`, since the default number of teams is `40`.  
- Removed `memcpy` and moved the `team_alltoall_world_dup` array to device memory.

tests/functional_tests/tester.cpp
```diff
@@ -628,33 +557,25 @@ void Tester::barrier() {
...
-  return cycles / gpu_frequency_MHz;
+double Tester::gpuCyclesToMicroseconds(uint64_t cycles) {
+  return static_cast<double>(cycles) / (wall_clk_rate * 1e-3);
 }
 
-uint64_t Tester::timerAvgInMicroseconds() {
-  uint64_t sum = 0;
+double Tester::timerAvgInMicroseconds() {
+  double sum = 0;
+  min_start_time = 0xFFFFFFFFFFFFFFFF;
```
BKP, line 566: UINT64_MAX in cstdint.h

tests/functional_tests/tester.cpp
```diff
@@ -527,32 +468,10 @@ void Tester::execute() {
     // data validation
```
BKP, line 468: Please remove the TODO on line 440. The timer type is "long long int" for wall_clock64. It's not technically a uin64_t type (we're depending on cstdint.h defining it as long long int); however, I think that it's a reasonable assumption in rocm.

Also, please remove the dump_stats and reset_stats lines on 460 and 461.

tests/functional_tests/tester.cpp
BKP, line 566: UINT64_MAX in cstdint.h

tests/functional_tests/tester.cpp
```diff
@@ -583,33 +502,43 @@ void Tester::print(uint64_t size) {
...
+  uint64_t total_size = size * num_timed_msgs;
   uint64_t timer_avg = timerAvgInMicroseconds();
   double latency_avg = static_cast<double>(timer_avg) / num_timed_msgs;
   double avg_msg_rate = num_timed_msgs / (timer_avg / 1e6);
 
   float total_kern_time_ms;
   CHECK_HIP(hipEventElapsedTime(&total_kern_time_ms, start_event, stop_event));
   float total_kern_time_s = total_kern_time_ms / 1000;
+
+  double time_ms = gpuCyclesToMicroseconds(max_end_time - min_start_time);
```
BKP, line 517: time_us

tests/functional_tests/tester.cpp
```diff
@@ -583,33 +502,43 @@ void Tester::print(uint64_t size) {
     return;
   }
 
+  /**
+   * Calculate total amount of data transfered
+   */
+  uint64_t total_size = size * num_timed_msgs;
   uint64_t timer_avg = timerAvgInMicroseconds();
   double latency_avg = static_cast<double>(timer_avg) / num_timed_msgs;
```
BKP, line 510: Can you print min,max,avg,std_dev for latency? I think that it's straightforward to do since you're capturing the timers for each participant.

It might help catch errors if something other than avg is reported. For instance, min should never be 0 in reality.

tests/functional_tests/alltoall_tester.cpp
```diff
@@ -52,30 +50,32 @@ ALLTOALL_DEF_GEN(unsigned int, uint)
...
+  dest_buf += wg_id * n_pes * size;
+
   __syncthreads();
 
-  uint64_t start;
   for (int i = 0; i < loop + skip; i++) {
     if (i == skip && hipThreadIdx_x == 0) {
-      start = rocshmem_timer();
+      start_time[wg_id] = wall_clock64();
     }
```
BKP, line 77: Can rocshmem_timer just be removed from the library completely? I don't see it deleted in this changeset.

It used to be that there was not a wall_clock64 wrapper for s_memrealtime (so we had to implement a function for it).

tests/functional_tests/primitive_tester.cpp
```diff
@@ -100,20 +108,21 @@ PrimitiveTester::~PrimitiveTester() {
...
   hipLaunchKernelGGL(PrimitiveTest, gridSize, blockSize, shared_bytes, stream,
-                     loop, args.skip, timer, s_buf, r_buf, size, _type,
-                     _shmem_context);
+                     loop, args.skip, start_time, end_time, s_buf, r_buf,
+                     size, _type, _shmem_context);
 
-  num_msgs = (loop + args.skip) * gridSize.x;
-  num_timed_msgs = loop;
+  num_msgs = (loop + args.skip) * gridSize.x * blockSize.x;
+  num_timed_msgs = loop * gridSize.x * blockSize.x;
```
BKP, line 125: Yes, this looks right.

tests/functional_tests/primitive_tester.cpp
```diff
@@ -100,20 +108,21 @@ PrimitiveTester::~PrimitiveTester() {
 }
 
 void PrimitiveTester::resetBuffers(uint64_t size) {
-  memset(s_buf, '0', args.max_msg_size * args.wg_size);
-  memset(r_buf, '1', args.max_msg_size * args.wg_size);
+  size_t buff_size = size * args.wg_size * args.num_wgs;
+  memset(s_buf, '0', buff_size);
```
BKP, line 112: Something to be aware of here is that there is same_value compression in some parts of our data fabric. Sending a block with repeating value might get compressed down into a bit_stream that transfers more quickly than a normal data stream.

I think that some other members of your team have run into this problem before. I was not aware of this data fabric optimization until recently. The easy solution is to iota value the bit_stream, but it would require changes to the validation functions. :(

Bottom line is that the reported latency/bandwidth may be **optimistic** when compared with real workloads. It's most likely to get reported (or surface externally) on the primitive tester results.

tests/functional_tests/primitive_tester.cpp
```diff
@@ -79,7 +86,7 @@ __global__ void PrimitiveTest(int loop, int skip, uint64_t *timer, char *s_buf,
   __syncthreads();
```
BKP, line 86: I don't have time to do it, but you might double check that the __syncthreads is present on all the "normal" variants of these tests where you protect the timer result with hipThreadIdx_x == 0.

It's important for __syncthreads to be there before the write on these tests, but __syncthreads should not be there before the timer writes on the warp variants of these tests.

tests/functional_tests/tester.cpp
```diff
@@ -628,33 +557,25 @@ void Tester::barrier() {
...
 
-  /**
-   * TODO: (bpotter/avinash) Modify the calcuation for the Tiled version of
-   *       puts and gets at wavefront level
-  */
-  for (uint64_t i = 0; i < args.num_wgs; i++) {
+  for (uint64_t i = 0; i < num_timers; i++) {
+    timer[i] = end_time[i] - start_time[i];
     sum += gpuCyclesToMicroseconds(timer[i]);
+    min_start_time = (start_time[i] < min_start_time)
```
BKP, line 572: Can you break these out into min,max,std_dev?

tests/functional_tests/tester.hpp
```diff
@@ -27,6 +27,7 @@
 #include <vector>
 
 #include "tester_arguments.hpp"
+#include "../src/util.hpp"
```
BKP, line 30: I don't think that you need this.

avinashkethineedi:
Closing this PR; it will be divided into multiple commits and separate PRs.

PR #32, opened 2025-02-03
[IPC] Fix ROCSHMEM_SIGNAL_ADD

Yiltan:
There was a typo where `ROCSHMEM_SIGNAL_ADD`  was applying the atomic operation to the wrong address


avinashkethineedi (approved):
Looks good to me.  

We should update the [signaling operations tests](https://github.com/ROCm/rocSHMEM/blob/develop/tests/functional_tests/signaling_operations_tester.cpp) to verify the values in `sig_addr` after the `putmem_signal*` API transfers the data, once PRs #30 and #32 are merged

PR #34, opened 2025-02-04
Fix Signalling Operations Functional Test

Yiltan:
Fixes a bug where: rocshmem_signal_fetch_wave is called on the global context but we would wait on a private context rocshmem_ctx_quiet(ctx);

- Split existing kernel into two tests that evaluate:
  1) rocshmem_ctx_putmem_signal*
  2) rocshmem_signal_fetch*

### DIFF
--- a/tests/functional_tests/signaling_operations_tester.cpp
+++ b/tests/functional_tests/signaling_operations_tester.cpp
@@ -201,7 +201,6 @@ void SignalingOperationsTester::verifyResults(uint64_t size) {
       } else if (ROCSHMEM_SIGNAL_ADD == sig_op) {
         uint64_t value = *sig_addr;
         uint64_t expected_value = (args.myid + 123); // Initial Value
-        uint64_t num_waves = 1;
 
         switch (_type) {
           case PutSignalTestType:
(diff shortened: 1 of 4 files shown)


Yiltan:
@avinashkethineedi The requested updates to the putmem_signal tests should now be present

tests/functional_tests/signaling_operations_tester.cpp
```diff
@@ -120,44 +150,83 @@ void SignalingOperationsTester::launchKernel(dim3 gridSize, dim3 blockSize, int
...
+          case PutSignalNBITestType:
+            expected_value += ((args.skip + args.loop) * args.wg_size * args.num_wgs);
+            break;
+          case WGPutSignalTestType:
+          case WGPutSignalNBITestType:
+            expected_value += ((args.skip + args.loop) * args.num_wgs);
+            break;
+          case WAVEPutSignalTestType:
+          case WAVEPutSignalNBITestType:
+            num_waves = max(1, (args.num_wgs / __AMDGCN_WAVEFRONT_SIZE__));
```
avinashkethineedi: - `args.wg_size` instead of `args.num_wgs`
- `num_waves` is already calculated in [`tester.cpp`](https://github.com/ROCm/rocSHMEM/blob/8d74c7b73e39da89d39f09a02795bf5711f2354c/tests/functional_tests/tester.cpp#L61) file, it would be better to use that

Yiltan: Applied: https://github.com/ROCm/rocSHMEM/pull/34/commits/b83ff2fa84d82949dcf2e9724513836a1bd482d9

avinashkethineedi (approved):

PR #37, opened 2025-02-05
locked specific version on ompi and ucx

akolliasAMD:


scripts/install_dependencies.sh
```diff
@@ -42,8 +43,10 @@ make install
 # Step 2: Install OpenMPI with UCX support
 cd $_DEPS_SRC_DIR
 rm -rf ompi
-git clone --recursive $_OMPI_REPO -b $_OMPI_BRANCH
+git clone --recursive $_OMPI_REPO
```
Yiltan, line 46: LGTM

`--recursive` could probably be removed here to speed up the script but its not necessary

Yiltan (approved):

scripts/install_dependencies.sh
edgargabriel: ```--recursive``` is required, otherwise Open MPI will not compile. The ```--recursive``` makes sure that the git submodules used by Open MPI will also be checked out along (e.g. pmix).

scripts/install_dependencies.sh
```diff
@@ -42,8 +43,10 @@ make install
...
 cd ompi
+git checkout $_OMPI_COMMIT_HASH
+git submodule update --init --recursive
```
edgargabriel, line 49: Well, let me correct myself, since we do the ```submodule update``` here, @Yiltan is correct that the ```--recursive``` might not be required. But it also shouldn't cause any issues.

edgargabriel (approved):

PR #49, opened 2025-03-04
Refactor RO backend data structures

avinashkethineedi:
- Removed `hdp` and `ipc` pointers from `BlockHandle`
- Aligned `RO stats` objects with number of `RO contexts`
- Added run commands for `rocshmem_g` and `rocshmem_p` API tests in `driver.sh`
- Allocated rocshmem API return buffers based on number of device contexts.
- Moved the status flag buffer allocation to the backend.
    - Associated the status flag address with each blocking call request to notify the GPU thread.
    - Removed dependency on `threadId` for determining the appropriate status flag index.

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -179,6 +179,20 @@ TestRMA() {
 
   ExecTest  "teamctxget"       2       1            1         1048576
 
+  ExecTest  "g"                2       1            1         1048576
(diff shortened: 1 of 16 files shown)


edgargabriel (approved):
Advocating to merge the PR and resolve any outstanding issues in subsequent PR, in order to advance the branch and enable multiple people to work/debug on it easier.

PR #51, opened 2025-03-14
Updated IPC detection logic

Yiltan:
- Added new environment variable `ROCSHMEM_RO_DISABLE_IPC`
    - Allows us to enable/disable IPC at runtime for RO backend
    - No longer require us to recompile to enable/disable
- Updated `isIpcAvailable(...)` logic
     - Previously required specific process mapping to provide correct results
     - New logic should work for arbitrary process mappings

### DIFF
--- a/src/util.hpp
+++ b/src/util.hpp
@@ -269,6 +269,13 @@ void rocm_memory_lock_to_fine_grain(void* ptr, size_t size, void** gpu_ptr,
 // Returns clock frequency used by s_memrealtime() in Mhz
 uint64_t wallClk_freq_mhz();
 
+struct rocshmem_env_config_t {
+  int ro_disable_ipc = 0;
+};
+extern struct rocshmem_env_config_t rocshmem_env_config;
+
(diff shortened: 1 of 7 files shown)


src/util.hpp
```diff
@@ -269,6 +269,13 @@ void rocm_memory_lock_to_fine_grain(void* ptr, size_t size, void** gpu_ptr,
 // Returns clock frequency used by s_memrealtime() in Mhz
 uint64_t wallClk_freq_mhz();
 
+struct rocshmem_env_config_t {
+  int ro_disable_ipc = 0;
```
Yiltan, line 273: This struct is overkill for this specific feature but it could be good to add all new environment variables in a single place moving forward.

edgargabriel (approved):

src/util.hpp
edgargabriel: I think this makes sense, its an approach used by a number of libraries (including RCCL and UCX). This way we don't have to do getenv() throughout the code, but just once at one location

PR #52, opened 2025-03-17
Sync Reverse Offload Scripts

Yiltan:
- Disable IPC unit tests when IPC is not available in the rocSHMEM configuration

### DIFF
--- a/scripts/build_configs/ro_net
+++ b/scripts/build_configs/ro_net
@@ -19,12 +19,13 @@ cmake \
     -DPROFILE=OFF \
     -DUSE_GPU_IB=OFF \
     -DUSE_DC=OFF \
+    -DUSE_IPC=OFF \
     -DUSE_COHERENT_HEAP=ON \
+    -DUSE_THREADS=OFF \
+    -DUSE_WF_COAL=OFF \
+    -DUSE_HOST_SIDE_HDP_FLUSH=OFF\
     -DUSE_MANAGED_HEAP=OFF \
     -DUSE_RO=ON \
-    -DUSE_IPC=ON \
-    -DUSE_THREADS=ON \
-    -DUSE_WF_COAL=OFF \
     $src_path
 cmake --build . --parallel 8
 cmake --install .
--- a/scripts/build_configs/ro_net_debug
+++ b/scripts/build_configs/ro_net_debug
@@ -20,9 +20,12 @@ cmake \
     -DUSE_GPU_IB=OFF \
     -DUSE_DC=OFF \
     -DUSE_IPC=OFF \
-    -DUSE_THREADS=ON \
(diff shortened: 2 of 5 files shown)


scripts/build_configs/ro_net
```diff
@@ -19,12 +19,13 @@ cmake \
     -DPROFILE=OFF \
     -DUSE_GPU_IB=OFF \
     -DUSE_DC=OFF \
+    -DUSE_IPC=OFF \
     -DUSE_COHERENT_HEAP=ON \
+    -DUSE_THREADS=OFF \
+    -DUSE_WF_COAL=OFF \
```
avinashkethineedi, line 25: It would be better to turn this on

scripts/build_configs/ro_net
```diff
@@ -19,12 +19,13 @@ cmake \
...
+    -DUSE_HOST_SIDE_HDP_FLUSH=OFF\
```
avinashkethineedi, line 26: RO backend flushes HDP regularly, this needs to be turned on

scripts/build_configs/ro_net_debug
```diff
@@ -20,9 +20,12 @@ cmake \
     -DUSE_GPU_IB=OFF \
     -DUSE_DC=OFF \
     -DUSE_IPC=OFF \
-    -DUSE_THREADS=ON \
+    -DUSE_COHERENT_HEAP=ON \
+    -DUSE_THREADS=OFF \
     -DUSE_WF_COAL=OFF \
-    -DUSE_COHERENT_HEAP=OFF \
+    -DUSE_HOST_SIDE_HDP_FLUSH=OFF\
```
avinashkethineedi, line 26: HDP should be enabled for RO

scripts/build_configs/ro_net
Yiltan: This feature is currently disabled on `develop` , if we want to enable it, we should thoroughly test it, enable it, then create a separate PR.

scripts/build_configs/ro_net
Yiltan: This feature is currently disabled on `develop` , if we want to enable it, we should thoroughly test it, enable it, then create a separate PR.

scripts/build_configs/ro_net_debug
Yiltan: This feature is currently disabled on `develop` , if we want to enable it, we should thoroughly test it, enable it, then create a separate PR.

edgargabriel: I would tend to agree. If we have mostly tested (and shipped in 6.4) rocSHMEM with this feature being turned off, lets keep it this way as part of this pr. It is a separate commit/effort whether we want to turn it on.

edgargabriel (approved):

PR #58, opened 2025-03-20
RO/collectives: add linear algorithms using RPut/Rget

edgargabriel:
- make broadcast, alltoall and fcollect use a simple linear algorithm using MPI_RPut/Rget, but without blocking in the execution. This fixes a hang and blocking observed with previous implementations.
- remove the to_all interfaces, since they have been deprecated.
- remove the active-set interfaces, since they have been removed from rocSHMEM
- disable allocating the ata_buffer on the RO teams structure, since it is not used at the moment and saves as 128MB per team


src/reverse_offload/mpi_transport.cpp
```diff
@@ -469,568 +357,134 @@ void MPITransport::team_broadcast(void *dst, void *src, int size, int win_id,
...
-  MPI_Comm comm_cluster{
-      createComm(world_ranks[clust_id * clust_size], stride, clust_size)};
-  MPI_Comm comm_ring{createComm(world_ranks[new_rank % clust_size],
-                                stride * clust_size, num_clust)};
-
-  barrier(contextId, status, false, comm_cluster);
-  barrier(contextId, status, blocking, comm_ring);
+  NET_CHECK(MPI_Win_flush_all(bp->heap_window_info[win_id]->get_win()));
+  barrier(contextId, status, blocking, comm);
+  quiet(contextId,status);
```
avinashkethineedi:
```suggestion
  barrier(contextId, nullptr, false, comm);
  quiet(contextId, status);
```
to avoid notifying the GPU thread when barrier completes

avinashkethineedi (approved):

PR #59, opened 2025-03-21
Removed GPU_IB

Yiltan:


src/ipc/backend_ipc.cpp
```diff
@@ -35,7 +35,7 @@ namespace rocshmem {
 
 extern rocshmem_ctx_t ROCSHMEM_HOST_CTX_DEFAULT;
 
-rocshmem_team_t get_external_team(GPUIBTeam *team) {
```
Yiltan, line 38: @avinashkethineedi this is the only code change that wasn't an ifdef that was needed. I assume it's fine as the function just casts.

avinashkethineedi: Yes, it should be cast to `IPCTeam`, that was an error earlier.

edgargabriel:
This looks fundamentally good to me, but since it touches also a number of files in the ipc and ro codepath, the outcome of the CI is crucial to determine whether the PR breaks something accidentally.

edgargabriel (approved):

avinashkethineedi (approved):

PR #61, opened 2025-03-21
add rocshmem_barrier()

edgargabriel:
- add device side API for rocshmem_barrier (team)
- add implementation to both IPC and RO conduit
- cleanup some of the logic for sync, sync_all, barrier, barrier_all in the RO conduit
- add a testcase for the new function to the functional tests


tests/functional_tests/team_barrier_tester.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+TeamBarrierTester::~TeamBarrierTester() {
+  CHECK_HIP(hipFree(team_barrier_world_dup));
+}
+
+void TeamBarrierTester::preLaunchKernel() {
+  for (int team_i = 0; team_i < num_teams; team_i++) {
+    team_barrier_world_dup[team_i] = ROCSHMEM_TEAM_INVALID;
+    rocshmem_team_split_strided(ROCSHMEM_TEAM_WORLD, 0, 1, n_pes, nullptr, 0,
```
abouteiller, line 85: Would it make sense to test non-contiguous, non-world analogous teams, e.g. `0, 2, n_pes/2` and `1, 2, n_pes/2` (adjusted for n_pes parity)

abouteiller (approved):

tests/functional_tests/team_barrier_tester.cpp
edgargabriel: this would definitely make sense, I would suggest to add that as a follow-up item, since this should probably be done for more than just this test.

scripts/functional_tests/driver.sh
```diff
@@ -306,6 +307,7 @@ TestColl() {
   #       | Name             | Ranks | Workgroups | Threads | Max Message Size #
   ##############################################################################
   ExecTest  "barrierall"       2       1            1
+  ExecTest  "teambarrier"      2       1            1
```
avinashkethineedi, line 310: It would be better to add couple of multi work-groups tests, since each group operates on a different team.

edgargabriel: @avinashkethineedi and I had a chat offline, lets get this merged 'as-is', enhancing the collective CI is definitely necessary but we should do that as the next step. We will have to add more tests for larger number of pes (definitely more than 2), more threads, more workgroups, different team composition patterns. This should be done however for all collectives, not just this one new function

PR #62, opened 2025-03-24
Add AtomicWFQueue implementation and tests

avinashkethineedi:
- [feat: Add AtomicWFQueue implementation](https://github.com/ROCm/rocSHMEM/commit/ec3eb2e2f62dadb2631a3a90200301e91c39029d)
- [test: Add GTest for AtomicWFQueue](https://github.com/ROCm/rocSHMEM/commit/0e32808eba52fed82b24428234095948aea390f5)
- `AtomicWFQueue` will be used for the status flag, g return, and atomic return buffers of the default context.

### DIFF
--- a/src/containers/atomic_wf_queue.hpp
+++ b/src/containers/atomic_wf_queue.hpp
@@ -102,8 +102,7 @@ class AtomicWFQueue {
    *
    * This function inserts the specified value at the position indicated by 
    * the `tail_` of the AtomicWFQueue and increases the AtomicWFQueue size
-   * by one. If the AtomicWFQueue is full, the function waits until space
-   * becomes available. The enqueue operation follows a first-come, first-serve
+   * by one. The enqueue operation follows a first-come, first-serve
    * execution order.
    *
    * @param val The value to be inserted into the AtomicWFQueue.
@@ -127,7 +126,7 @@ class AtomicWFQueue {
    * @brief Inserts a new element at the end of the AtomicWFQueue.
    *
    * This function adds the specified value to the end of the AtomicWFQueue,
-   * updating the `tail_` and `size` accordingly. It is intended for
+   * updating the `tail_` and `curr_size_` accordingly. It is intended for
    * initializing the AtomicWFQueue with initial values.
    *
    * @note This function is not thread-safe and should only be used during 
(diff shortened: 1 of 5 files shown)


src/containers/atomic_wf_queue.hpp
```diff
@@ -0,0 +1,359 @@
...
+  }
+
+ private:
+
+  __device__ int atomic_load(const int* address) {
+    return __hip_atomic_load(address, __ATOMIC_SEQ_CST,
+                             __HIP_MEMORY_SCOPE_AGENT);
+  }
+
+  __device__ void atomic_store(int* address, int val) {
```
edgargabriel: NIT: since you used `const int* address` above for the atomic load, shouldn't it also be `const int val` here?

src/containers/atomic_wf_queue.hpp
```diff
@@ -0,0 +1,359 @@
+/******************************************************************************
+ * Copyright (c) 2024 Advanced Micro Devices, Inc. All rights reserved.
```
edgargabriel, line 2: NIT: need to determine whether we need to update the year number to 2025

edgargabriel (dismissed):

src/containers/atomic_wf_queue_impl.hpp
```diff
@@ -0,0 +1,164 @@
...
+     * operation. Ensures a first-come, first-serve execution order
+     */
+    TicketLockGuard<MutexType> guard(*enqueue_mutex_.get());
+
+    /**
+     * The queue is full; wait until space becomes available.
+     * Ideally, there should always be space available.
+     * If the queue is full, it indicates an unexpected issue.
+     */
+    while (is_full()) {}
```
BKP: This should probably be an error state rather than a wait.  It means that something is broken and we should error out.

Yiltan (dismissed):

src/containers/atomic_wf_queue_impl.hpp
avinashkethineedi: Yes, I included that to check for any unknown edge cases and to see if the while loop body executes twice. So far, in testing, I haven't seen that while loop running more than once. I'll replace it with the `assert`.

edgargabriel (approved):

PR #71, opened 2025-04-01
Substitute pow2bin allocator with a dlmalloc based allocator

abouteiller:
* imports dlmalloc (latest version MIT licensed)
* create an encapsulation class DLMalloc that exposes only relevant functionalities, to prevent using non-static/templated members of parent class we use the mspace variant of dlmalloc
* use DLMalloc in a new ShmemAllocatorStrategy
* replace pow2bin in single_heap

**Possible drawbacks/missing features:**
* [x] the MORECORE functionality is not implemented (would cause dependency between non-static class members and static functions) -> we cannot resize the symmetric heap (initial allocation only) **won't do for now**
* [x] MSPACE allocator stores some metadata in the symmetric heap, that means that metadata in device memory is manipulated from the host when doing the allocation. **ok for now**
* [x] I have set the alignment to 2MB, this may be an overkill in some cases, maybe using the memalign functionality of dlmalloc would be a better approach. **(now inherit alignment from the ALIGNMENT compile constant), adding full memalign support could be done in a followup**
* [x] Performance evaluation (verified on hyd7 IPC, functional tests show same results)
* [x] Unit testing
* [x] Error in team_ctx_infra_tester
* [x] Error in pingpong


src/ipc/backend_ipc.cpp
```diff
@@ -22,6 +22,7 @@
 
 #include "backend_ipc.hpp"
 #include "ipc_team.hpp"
+#include <iostream>
```
BKP: minor nit: please strip this out

BKP (commented):
It should have some unit tests just to kick the tires before we merge into the code.

src/memory/dlmalloc.cpp
```diff
@@ -0,0 +1,6359 @@
...
+#endif  /* ONLY_MSPACES */
+#endif  /* MSPACES */
+#ifndef MALLOC_ALIGNMENT
+#define MALLOC_ALIGNMENT ((size_t)(2 * sizeof(void *)))
+#endif  /* MALLOC_ALIGNMENT */
+#ifndef FOOTERS
+#define FOOTERS 0
+#endif  /* FOOTERS */
+#ifndef ABORT
+#define ABORT  abort()
```
Yiltan, line 675: Should this be `MPI_Abort(...)`, or does calling `abort()` make no difference? i.e does it matter for the `USAGE_ERROR_ACTION` macro?

edgargabriel:
@abouteiller to follow up on the comment on the JIRA ticket, is there a test (or at least visual confirmation by reading the code) that dlmalloc is able to combine two freed memory allocation stemming from separate rocshmem_alloc/free() operations if they end-up being consecutive in the memory? This was ultimately why we wanted to replace the previous allocator

abouteiller:
> @abouteiller to follow up on the comment on the JIRA ticket, is there a test (or at least visual confirmation by reading the code) that dlmalloc is able to combine two freed memory allocation stemming from separate rocshmem_alloc/free() operations if they end-up being consecutive in the memory? This was ultimately why we wanted to replace the previous allocator

Yes this is visible in mspace_free:5706 and on, freed chunks are consolidated with preceding, suceeding free space. It also look for a best-fit chunk in the gaps for allocating chunks (with 2 different strategies for small and large requests).

src/memory/dlmalloc.hpp
```diff
@@ -0,0 +1,156 @@
...
+                        [[maybe_unused]] size_t request_size) override {}
+
+  /**
+   * @brief Frees memory from the heap
+   *
+   * Released memory is tracked by bookkeeping structures within this class.
+   *
+   * @param[in] Raw pointer to heap memory
+   *
+   * @note We do not attempt to combine records together into larger
```
BKP: This comment is wrong based on how you your verbal description of the dlmalloc code.

BKP (dismissed):

src/ipc/backend_ipc.cpp
abouteiller: required by `175: std::cerr << "MPI_THREAD_MULTIPLE support disabled.\n";` as we do not inherint iostream include from pow2bins...binner.hpp

src/memory/dlmalloc.cpp
abouteiller: single node abort gets propagated by prte/mpich

abouteiller:
Passes all tests now, ready for review

Yiltan (approved):
LGTM - tested locally and all tests passed

edgargabriel (approved):

PR #72, opened 2025-04-01
Update `Barrier_All` and `Sync_All` APIs

avinashkethineedi:
* Added `barrier_all` and `sync_all` APIs at thread, wavefront, and workgroup levels
   - Added in both `IPC` and `RO` conduits
* Updated tests for `barrier_all` APIs
* Added new tests for `sync_all` APIs


scripts/functional_tests/driver.sh
```diff
@@ -303,11 +313,38 @@ TestColl() {
...
 
   ExecTest  "sync"             2       1            1
 
-  ExecTest  "syncall"          2       1            1
+  ExecTest  "syncall"       2       1            1
+  ExecTest  "syncall"       2       16           64
+  ExecTest  "syncall"       2       32           256
+  ExecTest  "syncall"       2       64           1024
+
+  ExecTest  "wavesyncall"   4       1            1
```
Yiltan: Is `4` here a typo? syncall, wgsyncall are only tested with two ranks

I'm not apposed to it, I think we should be testing with more than two ranks, just seems inconsistent

avinashkethineedi: Not necessarily, I was checking whether those tests pass with multiple PEs, and they do. But I'll remove it for now and add multi-PE support for all the collective tests in a separate PR

include/rocshmem/rocshmem.hpp
```diff
@@ -490,6 +490,36 @@ __device__ int rocshmem_team_translate_pe(rocshmem_team_t src_team,
...
+
+__device__ ATTR_NO_INLINE void rocshmem_barrier_all();
+
+/**
+ * @brief perform a collective barrier between all PEs in the system.
+ * The caller is blocked until the barrier is resolved.
+ *
+ * This function must be called as a wave-front collective.
+ *
+ * @param[in] handle GPU side handle.
```
edgargabriel, line 514: mismatch in argument name (handle vs. ctx)

include/rocshmem/rocshmem.hpp
```diff
@@ -515,6 +545,42 @@ __device__ ATTR_NO_INLINE void rocshmem_wg_barrier_all();
...
+ * @brief registers the arrival of a PE at a barrier.
+ * The caller is blocked until the synchronization is resolved.
+ *
+ * In contrast with the shmem_barrier_all routine, shmem_sync_all only ensures
+ * completion and visibility of previously issued memory stores and does not
+ * ensure completion of remote memory updates issued via OpenSHMEM routines.
+ *
+ * This function must be invoked by a single thread within the PE.
+ *
+ * @param[in] handle GPU side handle.
```
edgargabriel, line 558: same here

include/rocshmem/rocshmem.hpp
```diff
@@ -515,6 +545,42 @@ __device__ ATTR_NO_INLINE void rocshmem_wg_barrier_all();
...
+ * @brief registers the arrival of a PE at a barrier.
+ * The caller is blocked until the synchronization is resolved.
+ *
+ * In contrast with the shmem_barrier_all routine, shmem_sync_all only ensures
+ * completion and visibility of previously issued memory stores and does not
+ * ensure completion of remote memory updates issued via OpenSHMEM routines.
+ *
+ * This function must be called as a wave-front collective.
+ *
+ * @param[in] handle GPU side handle.
```
edgargabriel, line 576: and here

edgargabriel (approved):
LGTM, except for the few comments on the function documentaiton

include/rocshmem/rocshmem.hpp
avinashkethineedi: handle is consistently referred to as ctx across all APIs in that file

edgargabriel: in that case you can leave it as is, but that is unfortunate. I don't necessarily care about the term we use in the description, but the argument name should actually match what is in the function prototype listed below

edgargabriel: we might actually want to fix that throughout the header file in a subsequent PR

PR #73, opened 2025-04-02
Update `Barrier` and `Sync` APIs

avinashkethineedi:
* Added `barrier` and `sync` APIs at thread, wavefront, and workgroup levels
  * Added in both `IPC` and `RO` conduits
* Updated tests for `barrier` APIs
* Updated new tests for `sync` APIs
* Removed collective operations on default context


include/rocshmem/rocshmem.hpp
```diff
@@ -533,17 +529,47 @@ __device__ ATTR_NO_INLINE void rocshmem_wave_barrier_all();
...
+ *
+ * This function must be invoked by a single thread within the PE.
+ *
+ * @param[in] handle GPU side handle.
+ *
+ * @param[in] team The team on which to perform barrier synchronization
+ *
+ * @return void
+ */
+__device__ void rocshmem_barrier(rocshmem_ctx_t ctx, rocshmem_team_t team);
```
abouteiller: 1. We have a naming problem here: 
`shmem_barrier` is a pe-stride based collective, not a team based collective. (note that pe-stride barrier is deprecated as-of 1.6). An appropriate name would be `rocshmem_team_barrier`

2. Do we even have to provide these variants? In oshmem1.6 pp.104 we find the following note: 
> No team-based barrier is provided by OpenSHMEM, as a team may have any number of communication contexts associated with the team. Applications seeking such an idiom should call `shmem_ctx_quiet` on the desired communication context, followed by a call to `shmem_team_sync` on the desired team.

include/rocshmem/rocshmem.hpp
```diff
@@ -597,7 +619,41 @@ __device__ ATTR_NO_INLINE void rocshmem_wave_sync_all();
...
+ * ensure completion of remote memory updates issued via OpenSHMEM routines.
+ *
+ * This function must be invoked by a single thread within the PE.
+ *
+ * @param[in] handle GPU side handle.
+ * @param[in] team  Handle of the team being synchronized
+ *
+ * @return void
+ */
+__device__ ATTR_NO_INLINE void rocshmem_ctx_team_sync(
```
abouteiller, line 637: The oshmem version is named `shmem_team_sync` and doesn't take the context as parameter. 

In general, when a procedure takes a context as input it doesn't need the team, because every shmem_ctx_t context is associated with a team when created.

include/rocshmem/rocshmem.hpp
avinashkethineedi: As discussed internally, updated the API naming
commit: https://github.com/ROCm/rocSHMEM/pull/73/commits/7d41e0d4da0e2008c3c21ef778ab1faf7644b8d2

abouteiller (approved):

PR #75, opened 2025-04-09
Cleanup unused code in repository

BKP:
Several experimental/incomplete features were included in repository.

These features should be removed for public releases.


abouteiller (approved):

abouteiller:
dev_mono_linear is used only from slab_heap (which is removed here)

abouteiller:
Since Brandon is very busy atm I recommend we merge as-is and I will bring up a followup to remove dev_mono_linear

edgargabriel:
> Since Brandon is very busy atm I recommend we merge as-is and I will bring up a followup to remove dev_mono_linear

I would be ok with that

src/reverse_offload/backend_ro.cpp
```diff
@@ -242,7 +242,12 @@ void ROBackend::dump_backend_stats() {
     total += globalStats.getStat(i);
   }
 
-  uint64_t gpu_frequency_mhz{wallClk_freq_mhz()};
+  int device_id;
+  hipDeviceProp_t device_props;
+  CHECK_HIP(hipGetDevice(&device_id));
+  int wallClockMhz;
+  CHECK_HIP(hipDeviceGetAttribute(&wallClockMhz, hipDeviceAttributeWallClockRate, device_id));
```
avinashkethineedi, line 249: Moving this into backend constructor would be better since we already invoke device properties to populate few variables

PR #79, opened 2025-04-11
Update backend to use provided MPI communicator during library initialization

avinashkethineedi:
Updated backend to use the MPI communicator passed during the library initialization. If no communicator is explicitly provided, it defaults to `MPI_COMM_WORLD`.

### DIFF
--- a/src/backend_bc.cpp
+++ b/src/backend_bc.cpp
@@ -33,7 +33,15 @@
 
 namespace rocshmem {
 
-Backend::Backend() {
+#define NET_CHECK(cmd)                                       \
+  {                                                          \
+    if (cmd != MPI_SUCCESS) {                                \
+      fprintf(stderr, "Unrecoverable error: MPI Failure\n"); \
+      abort() ;                                              \
+    }                                                        \
+  }
+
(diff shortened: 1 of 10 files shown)


edgargabriel (dismissed):

src/backend_bc.cpp
```diff
@@ -33,7 +33,15 @@
 
 namespace rocshmem {
 
-Backend::Backend() {
+#define NET_CHECK(cmd)                                       \
+  {                                                          \
+    if (cmd != MPI_SUCCESS) {                                \
```
abouteiller, line 38: if we leave (or reset after the comm_dup, if we receive an external comm) the error handler to MPI_ERRORS_ARE_FATAL, it will kill the application in the MPI call if it were to return ~MPI_SUCCESS, making these extra checks redundant.

edgargabriel (approved):

src/backend_bc.cpp
Yiltan: Not for this PR but another comment on this macro. If we plan to remove it, we should remove elsewhere in the repo, we use it 2/3 times. If we plan to keep it, we should probably have on instance of it.

PR #80, opened 2025-04-13
Revamp the uniqueId code to support subgroups of processes

edgargabriel:
add new code for bootstrapping the uniqueId correctly.  The bootstrapping code has been extracted from the MSCCLPP library,
which in parts is based on the code from NVIDIA. The code has been modified to match the specific requirements of the rocSHMEM library.

This PR will require https://github.com/ROCm/rocSHMEM/pull/79 in order to work correctly.

Some additional comments:
 - the TRACE and INFO macros in the bootstrap directory should be replaced by DPRINTF. However, if I include util.hpp from the src directory to use DPRINTF, I get some weird errors because of redefinition of the 'memcpy' function in there. We can resolve this issue later and remove the TRACE and INFO macros.
 -  there are a couple of new environment variables, and I was thinking about adding them to the rocshmem_env_config structure. The problem however is that these environment variables are being used before the library is initialized, and hence it is unclear on how much value it would add to add them to the structure.

### DIFF
--- a/examples/rocshmem_init_attr_test.cc
+++ b/examples/rocshmem_init_attr_test.cc
@@ -30,24 +30,35 @@ using namespace rocshmem;
 
 int main (int argc, char **argv)
 {
-    int rank, nranks;
+    int world_rank, world_nranks;
     int ret;
     rocshmem_uniqueid_t uid;
     rocshmem_init_attr_t attr;
 
     MPI_Init(&argc, &argv);
-    MPI_Comm_rank (MPI_COMM_WORLD, &rank);
-    MPI_Comm_size (MPI_COMM_WORLD, &nranks);
+    MPI_Comm_rank (MPI_COMM_WORLD, &world_rank);
+    MPI_Comm_size (MPI_COMM_WORLD, &world_nranks);
+
+    // Create two disjoint groups of processes, each
+    // one creating a unique rocshmem environment independet
+    // of the other group
+    MPI_Comm newcomm;
+    int color = world_rank %2;
+    int rank, nranks;
+
+    MPI_Comm_split(MPI_COMM_WORLD, color, world_rank, &newcomm);
+    MPI_Comm_rank (newcomm, &rank);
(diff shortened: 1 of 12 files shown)


edgargabriel:
a rare occurrence, even the RO tests seemed to have passed :-)

edgargabriel:
I can confirm that when running this PR on top of PR #79, the revamped version of rocshmem_init_attr_test, i.e. we can do the initialization of two disjoints groups within the same MPI_COMM_WORLD.

There is also no conflict when merging

examples/rocshmem_init_attr_test.cc
```diff
@@ -30,24 +30,35 @@ using namespace rocshmem;
...
     rocshmem_init_attr_t attr;
 
     MPI_Init(&argc, &argv);
-    MPI_Comm_rank (MPI_COMM_WORLD, &rank);
-    MPI_Comm_size (MPI_COMM_WORLD, &nranks);
+    MPI_Comm_rank (MPI_COMM_WORLD, &world_rank);
+    MPI_Comm_size (MPI_COMM_WORLD, &world_nranks);
+
+    // Create two disjoint groups of processes, each
+    // one creating a unique rocshmem environment independet
```
abouteiller:
```suggestion
    // one creating a unique rocshmem environment independent
```

src/rocshmem.cpp
```diff
@@ -102,7 +103,7 @@ rocshmem_ctx_t ROCSHMEM_HOST_CTX_DEFAULT;
 
 [[maybe_unused]] __host__ int rocshmem_init_attr(unsigned int flags,
 						 rocshmem_init_attr_t *attr) {
-  MPI_Comm comm = MPI_COMM_WORLD;
+  MPI_Comm comm;
```
abouteiller:
```suggestion
  MPI_Comm comm = MPI_COMM_NULL;
```

abouteiller (dismissed):

abouteiller (approved):

PR #82, opened 2025-04-14
Code Coverage

reboss:
In this PR, I have added a helper script to generate a code coverage report.  I updated the helper scripts such that by default, code coverage instrumentation is not enabled, but a  user can define the `CODE_COV` environment variable to `-DBUILD_CODE_COVERAGE=ON` in order to override.

I also made it so that by default, the helper build scripts will build to the current working directory, but can be overriden by setting the `BUILD_DIR` environment variable.

This will generate a code coverage report and both out put the report to the terminal and serves an html report with a python http.server.


scripts/build_configs/ipc_single
```diff
@@ -50,6 +52,7 @@ cmake \
     -DUSE_SINGLE_NODE=ON \
     -DUSE_HOST_SIDE_HDP_FLUSH=OFF \
     -DBUILD_LOCAL_GPU_TARGET_ONLY=OFF \
+    -DCMAKE_C_COMPILER="/opt/rocm/bin/amdclang" \
```
Yiltan: Could this line be added to the CMakefile itself? Same goes for other build_configs

I think it will be around here: https://github.com/ROCm/rocSHMEM/blob/develop/CMakeLists.txt#L78

scripts/build_configs/codecov
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+export CODE_COV="-DBUILD_CODE_COVERAGE=ON"
+
+src_path=$(dirname "$(realpath $0)")
+
+BUILD_DIR="$PWD/ipc" $src_path/ipc_single
```
Yiltan: Could we do something like:

```suggestion
mkdir -p $PWD/ipc" 
cd $PWD/ipc" 
$src_path/ipc_single
cd ..
```

(I think there is cleaner way to do it, so feel free to do something different)

See my comment in `build_config/ipc` for my reasoning

scripts/build_configs/ipc_single
```diff
@@ -36,6 +36,8 @@ fi
 src_path=$(dirname "$(realpath $0)")/../../
 
 cmake \
+    -B${BUILD_DIR:-.} \
```
Yiltan: In our `install_dependencies.sh` script we use `BUILD_DIR` to fit in with ROCm-CI. I am concerned this could cause issues as we have the same variable name but for a different use

README.md
```diff
@@ -154,6 +154,20 @@ To run the tests, you may use the driver scripts provided in the `./scripts/` di
 ./scripts/unit_tests/driver.sh ./build/tests/unit_tests/rocshmem_unit_tests all
 ```
 
+## Code Coverage
```
Yiltan, line 157: I assume this doesn't need to be added to the `docs/` folder as this is for us and not something that we'd expect rocSHMEM users/customers to do?

scripts/build_configs/codecov
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+export CODE_COV="-DBUILD_CODE_COVERAGE=ON"
```
Yiltan: Would this not evaluate to the following, in the build_configs?

```
-DBUILD_CODE_COVERAGE=-DBUILD_CODE_COVERAGE=ON
```

README.md
reboss: Probably not something that most customers would want to do.  More of a developer/contributor thing.

scripts/build_configs/codecov
reboss: You're right haha.  I mean, it works, but yea I'll update it to just "ON"

scripts/build_configs/ipc_single
reboss: Gotcha, I changed it so that I am just creating and changing the current working directory for each target

Yiltan (approved):

PR #83, opened 2025-04-15
bugfix: do not dereference ctx during create_ctx if we did run out of ctx

abouteiller:
This scenario happens when the functional test teamctxinfra is invoked in `functional_tests/drivers.sh teamctxinfra`, that invokes the teamctxinfra tester with a different setup than `all` would do

` mpirun -n 2 -mca pml ucx -mca osc ucx -x ROCSHMEM_MAX_NUM_CONTEXTS=1 -x UCX_ROCM_IPC_SIGPOOL_MAX_ELEMS=16384 --map-by numa --timeout 300 tests/functional_tests/rocshmem_example_driver -a 38 -w 1 -z 1 -s 8`

notable difference is that `all` would pass `-x ROCSHMEM_MAX_NUM_CONTEXTS=1024`

This unexpected setup would exercise a bug in backends `create_ctx` where we would dereference `ctx->SetFence` during create when `ctx` is `nullptr`. 

In addition the ctxinfra tester would push back the nullptr ctx from the prior failed create into the freelist when calling free_ctx with a nullptr param, which would explode in an even more bizare way later, assert instead.


edgargabriel (approved):
LGTM, nice catch!

PR #89, opened 2025-04-22
unify env variables and use DPRINTF

edgargabriel:
this is a PR with two minor changes:
 - add the new environment variables introduced with the unique ID bootstrapping to the rocshmem_env_config environment. Also, convert this to a class instead of struct in order to be able to have the constructor run before rocshmem_init, since we need some of the values as part of the initialization process. This allows us to get rid of two files in the bootstrap directory
 - replace the TRACE and INFO macros in the bootstrap directory with DPRINTF to be more consistent with the rest of the code base,


Yiltan (dismissed):

src/util.cpp
```diff
@@ -121,21 +123,63 @@ void rocm_memory_lock_to_fine_grain(void* ptr, size_t size, void** gpu_ptr,
...
+  if (nullptr != env_value) {
+    bootstrap_timeout = atoi(env_value);
+  }
+
+  env_value = getenv("ROCSHMEM_HOSTID");
+  if (nullptr != env_value) {
+    hostid = std::string(env_value);
+  }
+
+  env_value = getenv("ROCSHMEM_SOCKET_FAMILY");
```
abouteiller: optional: change all these to `ROCSHMEM_BOOTSTRAP_SOCKET_FAMILY`, `BOOTSTRAP_SOCKET_IFNAME`, `ROCSHMEM_BOOTSTRAP_HOSTID`

abouteiller (dismissed):

src/util.cpp
edgargabriel: since CI is not yet done, I will implement the suggested namechanges, they make sense

abouteiller (approved):

PR #90, opened 2025-04-22
use correct MPI initialization method

edgargabriel:
rocSHMEM requires that the MPI library is initialized using THREAD_MULTIPLE support. Lets use that function therefore in our examples.

### DIFF
--- a/examples/rocshmem_init_attr_test.cc
+++ b/examples/rocshmem_init_attr_test.cc
@@ -61,6 +61,9 @@ int main (int argc, char **argv)
     int provided;
 
     MPI_Init_thread (&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
+    if (provided != MPI_THREAD_MULTIPLE) {
+      std::cerr << "MPI_THREAD_MULTIPLE support disabled.\n";
+    }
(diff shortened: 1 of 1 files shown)


examples/rocshmem_init_attr_test.cc
```diff
@@ -58,8 +58,9 @@ int main (int argc, char **argv)
     int ret;
     rocshmem_uniqueid_t uid;
     rocshmem_init_attr_t attr;
+    int provided;
 
-    MPI_Init(&argc, &argv);
+    MPI_Init_thread (&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
```
abouteiller, line 63:
```suggestion
    MPI_Init_thread (&argc, &argv, MPI_THREAD_MULTIPLE, &provided);
    if (provided != MPI_THREAD_MULTIPLE) {
      std::cerr << "MPI_THREAD_MULTIPLE support disabled.\n";
    }
```

Yiltan (approved):

abouteiller (approved):

avinashkethineedi (approved):

PR #86, opened 2025-04-15
Add address sanitizer build options

abouteiller:
Also a bit of documentation on how to debug

TODOs: 
* [x] compile with xnack+
* [x] document requirements for compiling/running with asan
* [ ] bug: code crashes when executed under asan

PR #94, opened 2025-04-29
Added XNACK support

Yiltan:
# Changes
- Updated CMakefiles so we have the following offload targets
    - `gfx90a:xnack-`
    - `gfx90a:xnack+`
    - `gfx942:xnack-`
    - `gfx942:xnack+`
- Updated examples code instruction to reflect this change 

# Reason
- ROCm-CI uses `xnack+/-` for GPU targets, so we want rocSHMEM to match that
- Kernels compiled with `gfx942` are not the same as those compiled with `gfx942:xnack-`
    - Same behaviour but different symbols, so causes linking issues if we are not explicit

# Notes
This may conflict with #86

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,10 @@ set(ROCMCHECKS_WARN_TOOLCHAIN_VAR OFF)
 include(cmake/rocm_local_targets.cmake)
 
 set(DEFAULT_GPUS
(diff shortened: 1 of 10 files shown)


edgargabriel (approved):
LGTM

abouteiller (approved):

PR #102, opened 2025-05-06
Remove unused code

BKP:
Removed several files that were not needed.

Instead of using inline assembly for uncached loads, dereference volatile pointers instead.


avinashkethineedi (approved):

tests/unit_tests/free_list_gtest.cpp
```diff
@@ -38,12 +38,12 @@ template <typename List, typename Value>
 __global__ void pop_all(List* list, Value* values, const std::size_t count) {
   const auto stride = blockDim.x * gridDim.x;
   const auto thread_index = blockIdx.x * blockDim.x + threadIdx.x;
-  // One push per block. block size is always WF_SIZE
-  for (std::size_t i = thread_index; i < count * WF_SIZE; i += stride) {
+  // One push per block. block size is always __AMDGCN_WAVEFRONT_SIZE
```
Yiltan, line 41: This can be addressed in a future PR

I think this is fine for the machine we test on, but if we use a machine with a new ROCm, I have a feeling it will cause issues. I think it's either ROCm 6.4 or 6.5

Yiltan (approved):

PR #104, opened 2025-05-06
[GDA] Fix ibv_reg_mr when using sub-communicators

Yiltan:
- Fails at ibv_reg_mr when using subcommunicators
  - Issue can be replicated with the `rocshmem_init_attr_test` test
- This is a result incorrect merge conflict resolution when cherry-picking 
  - #79 
  - The code has been adjusted to now work

### DIFF
--- a/src/memory/symmetric_heap.hpp
+++ b/src/memory/symmetric_heap.hpp
@@ -117,7 +117,7 @@ class SymmetricHeap {
   /**
    * @brief Implementation of remote heaps
    */
-  RemoteHeapInfoType remote_heap_info_{single_heap_.get_base_ptr(), single_heap_.get_size()};
+  RemoteHeapInfoType remote_heap_info_{};
 };
 
 }  // namespace rocshmem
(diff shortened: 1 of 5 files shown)


src/memory/symmetric_heap.hpp
```diff
@@ -117,7 +117,7 @@ class SymmetricHeap {
   /**
    * @brief Implementation of remote heaps
    */
-  RemoteHeapInfoType remote_heap_info_{single_heap_.get_base_ptr(), single_heap_.get_size()};
+  RemoteHeapInfoType remote_heap_info_{};
```
BKP, line 120: Why are the initialization fields removed?

Yiltan: See PR #79 from `develop`:

https://github.com/ROCm/rocSHMEM/commit/05755847f5791aae615b8eef2990eb371231bc56#diff-544c4f57cdc771e20a1822b5beb2b4d4b43f4fef4da81a254c0ab8f0b929da13

We want to pass a communicator (other than `MPI_COMM_WORLD`) to `RemoteHeapInfo` but we do not know which communicator will be passed until we call the constructor to `SymmetricHeap`. So we initialize `RemoteHeapInfoType` when we call the `SymmetricHeap` constructor

BKP (approved):

PR #107, opened 2025-05-07
Serialize on the queue pair interface with PE-based divergence

BKP:
The queue pair interface needs to be executed without wavefront divergence and it's not tolerant of buildings WQEs for different QPs within the same wavefront (at the same time).

To avoid having to rewrite the code, we're going to serialize the wavefronts such the threads within the wavefront will take a turn per PE.


BKP:
Closing - pushed patch from wrong branch.

PR #115, opened 2025-05-08
Refactor several classes and bugfixes

BKP:
Combine the backend, network, and connection classes into a new class "gda_device".

Enables the default context and integrates Avinash's changes.


src/gpu_ib/context_ib_device.cpp
```diff
@@ -26,29 +26,23 @@
...
+  barrier_sync = device->barrier_sync;
+  device->initialize_context(this, idx);
+  size_t barrier_sync_offset = idx * ROCSHMEM_BARRIER_SYNC_SIZE;
+  barrier_sync = device->barrier_sync + barrier_sync_offset;
 }
 
 __device__ void GPUIBContext::quiet() {
-  for (int k = 0; k < networkImpl.num_pes; k++) {
-    getQueuePair(k)->quiet();
+  for (int k = 0; k < num_pes; k++) {
```
avinashkethineedi, line 44: team specific context based quiet will loop over all the PEs

IPC: https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/context_ipc_device.cpp#L105

avinashkethineedi (approved):
LGTM, I can take care of the team-specific changes in my next PR, as I’m currently working on the team barrier.

src/gpu_ib/context_ib_device.cpp
BKP: I don't understand why this is relevant.

avinashkethineedi: it calls quiet on all the queue pairs when it doesn't have to

PR #124, opened 2025-05-12
Detailed logs

abouteiller:
Improve debug features by having parameters be printed with the DEBUG printfs

### DIFF
--- a/src/rocshmem_gpu.cpp
+++ b/src/rocshmem_gpu.cpp
@@ -283,14 +283,14 @@ __device__ Context *get_internal_ctx(rocshmem_ctx_t ctx) {
   return reinterpret_cast<Context *>(ctx.ctx_opaque);
 }
 
-__device__ int rocshmem_wg_ctx_create(long option, rocshmem_ctx_t *ctx) {
-  GPU_DPRINTF("Function: rocshmem_ctx_create\n");
+__device__ int rocshmem_wg_ctx_create(long options, rocshmem_ctx_t *ctx) {
+  GPU_DPRINTF("Function: rocshmem_wg_ctx_create (options=%ld)\n", options);
   bool result{true};
   if (get_flat_block_id() == 0) {
     ctx->team_opaque = reinterpret_cast<TeamInfo *>(ROCSHMEM_CTX_DEFAULT.team_opaque);
-    result = device_backend_proxy->create_ctx(option, ctx);
+    result = device_backend_proxy->create_ctx(options, ctx);
     if(result) {
-      reinterpret_cast<Context *>(ctx->ctx_opaque)->setFence(option);
+      reinterpret_cast<Context *>(ctx->ctx_opaque)->setFence(options);
     }
   }
   __syncthreads();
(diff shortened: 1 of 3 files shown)


src/rocshmem_gpu.cpp
```diff
@@ -283,23 +283,24 @@ __device__ Context *get_internal_ctx(rocshmem_ctx_t ctx) {
   return reinterpret_cast<Context *>(ctx.ctx_opaque);
 }
 
-__device__ int rocshmem_wg_ctx_create(long option, rocshmem_ctx_t *ctx) {
-  GPU_DPRINTF("Function: rocshmem_ctx_create\n");
+__device__ int rocshmem_wg_ctx_create(long options, rocshmem_ctx_t *ctx) {
+  GPU_DPRINTF("Function: rocshmem_ctx_create (options=%ld)\n", options);
```
Yiltan: Not introduced by this PR but

```suggestion
  GPU_DPRINTF("Function: rocshmem_wg_ctx_create (options=%ld)\n", options);
```

src/rocshmem_gpu.cpp
```diff
@@ -283,23 +283,24 @@ __device__ Context *get_internal_ctx(rocshmem_ctx_t ctx) {
...
   }
   __syncthreads();
   return result == true ? 0 : -1;
 }
 
 __device__ int rocshmem_wg_team_create_ctx(rocshmem_team_t team, long options,
-                                            rocshmem_ctx_t *ctx) {
-  GPU_DPRINTF("Function: rocshmem_team_create_ctx\n");
+                                           rocshmem_ctx_t *ctx) {
+  GPU_DPRINTF("Function: rocshmem_team_create_ctx (team=%zd, options=%ld)\n",
```
Yiltan:
```suggestion
  GPU_DPRINTF("Function: rocshmem_wg_team_create_ctx (team=%zd, options=%ld)\n",
```

src/rocshmem_gpu.cpp
```diff
@@ -321,139 +322,144 @@ __device__ int rocshmem_wg_team_create_ctx(rocshmem_team_t team, long options,
 
 __device__ void rocshmem_wg_ctx_destroy(
     [[maybe_unused]] rocshmem_ctx_t *ctx) {
-  GPU_DPRINTF("Function: rocshmem_ctx_destroy\n");
+  GPU_DPRINTF("Function: rocshmem_ctx_destroy (ctx=%zd)\n",
```
Yiltan:
```suggestion
  GPU_DPRINTF("Function: rocshmem_wg_ctx_destroy (ctx=%zd)\n",
```

src/util.hpp
```diff
@@ -62,9 +62,9 @@ namespace rocshmem {
 #endif
 
 #ifdef DEBUG
-#define GPU_DPRINTF(...)      \
-  do {                        \
-    gpu_dprintf(__VA_ARGS__); \
+#define GPU_DPRINTF(...)                                         \
+  do {                                                           \
+    gpu_dprintf("WG (%u, %u, %u) TH (%u, %u, %u) " __VA_ARGS__); \
```
Yiltan, line 67: Would it be worth adding which PE we are printing from? Not sure if this will be useful if we initialize rocSHMEM from a sub-communicator, since there will be multiple PEs with the same rank. So this comment is optional

```suggestion
    gpu_dprintf("[%d] WG (%u, %u, %u) TH (%u, %u, %u) " __VA_ARGS__); \
```

abouteiller: I thought about that, we have `mpiexec --output rank` to show the global pe-rank, we could show the caller my-pe in the team here?

Yiltan: I forgot about that flag. This probably isn't needed

Yiltan (approved):

edgargabriel (approved):

PR #125, opened 2025-05-13
Fix broken rocSHMEM 6.4.0 APIs

avinashkethineedi:
Fixes the following APIs
- `rocshmem_ctx_my_pe`
  - returns team relative PE numbering
- `rocshmem_ctx_n_pes`
  - return team relative size
- `rocshmem_ctx_wg_team_sync`
  - uses the correct synchronization buffers

### DIFF
--- a/src/ipc/context_ipc_device_coll.cpp
+++ b/src/ipc/context_ipc_device_coll.cpp
@@ -105,8 +105,9 @@ __device__ void IPCContext::sync(rocshmem_team_t team) {
   int pe_start = team_obj->tinfo_wrt_world->pe_start;
   int pe_stride = team_obj->tinfo_wrt_world->stride;
   int pe_size = team_obj->num_pes;
+  long *p_sync = team_obj->barrier_pSync;
 
-  internal_sync(pe, pe_start, pe_stride, pe_size, barrier_sync);
+  internal_sync(pe, pe_start, pe_stride, pe_size, p_sync);
 }
 
 __device__ void IPCContext::sync_all() {
--- a/src/rocshmem_gpu.cpp
+++ b/src/rocshmem_gpu.cpp
@@ -606,7 +606,8 @@ __device__ void rocshmem_wg_team_sync(rocshmem_team_t team) {
 __device__ int rocshmem_ctx_n_pes(rocshmem_ctx_t ctx) {
   GPU_DPRINTF("Function: rocshmem_n_pes\n");
 
-  return get_internal_ctx(ctx)->num_pes;
+  TeamInfo *tinfo = reinterpret_cast<TeamInfo *>(ctx.team_opaque);
+  return tinfo->size;
 }
 
 __device__ int rocshmem_n_pes() {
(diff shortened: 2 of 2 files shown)


src/rocshmem_gpu.cpp
```diff
@@ -616,7 +617,21 @@ __device__ int rocshmem_n_pes() {
 __device__ int rocshmem_ctx_my_pe(rocshmem_ctx_t ctx) {
   GPU_DPRINTF("Function: rocshmem_ctx_my_pe\n");
 
-  return get_internal_ctx(ctx)->my_pe;
+  TeamInfo *tinfo = reinterpret_cast<TeamInfo *>(ctx.team_opaque);
+  int my_pe{get_internal_ctx(ctx)->my_pe};
```
abouteiller, line 621: Why isn't get_internal_ctx(ctx)->my_pe already translated?

avinashkethineedi: All contexts gets created during the initialization phase ([setup_ctxs](https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/backend_ipc.cpp#L115)), where the variables `my_pe` and `n_pes` are also initialized as part of the context setup. These values corresponds to the team world since we won't be knowing what team they will be associated with and also some internal APIs are depended on those values. Since user contexts contain an [opaque team](https://github.com/ROCm/rocSHMEM/blob/develop/include/rocshmem/rocshmem_common.hpp#L112) object, retrieving `my_pe` and `n_pes` through them is more straightforward.

src/rocshmem_gpu.cpp
```diff
@@ -616,7 +617,21 @@ __device__ int rocshmem_n_pes() {
...
+  int pe_start{tinfo->pe_start};
+  int stride{tinfo->stride};
+  int size{tinfo->size};
+
+  int translated_pe = (my_pe - pe_start) / stride;
```
abouteiller, line 626: (edit: I proposed to use translate_pe from this file, but does the opposite translation team_pe->world_pe, so it is not duplicate for this snippet)
 
This however looks similar: https://github.com/ROCm/rocSHMEM/blob/67bff9ca308e7a68355b0399763824814b009717/src/team.hpp#L146

Yiltan (approved):

src/ipc/context_ipc_device_coll.cpp
```diff
@@ -105,12 +105,13 @@ __device__ void IPCContext::sync(rocshmem_team_t team) {
   int pe_start = team_obj->tinfo_wrt_world->pe_start;
   int pe_stride = team_obj->tinfo_wrt_world->stride;
   int pe_size = team_obj->num_pes;
+  long *p_sync = team_obj->barrier_pSync;
 
   internal_sync(pe, pe_start, pe_stride, pe_size, barrier_sync);
```
Yiltan: This uses p_sync in develop: 

https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/context_ipc_device_coll.cpp#L132

Not sure which is correct

src/ipc/context_ipc_device_coll.cpp
```diff
@@ -105,12 +105,13 @@ __device__ void IPCContext::sync(rocshmem_team_t team) {
...
 }
 
 __device__ void IPCContext::sync_all() {
-  internal_sync(my_pe, 0, 1, num_pes, barrier_sync);
+  internal_sync(my_pe, 0, 1, num_pes, p_sync);
```
Yiltan: This is unchanged in develop:

https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/context_ipc_device_coll.cpp#L160

PR #127, opened 2025-05-15
re-add code to select closest NIC to a GPU

edgargabriel:


BKP (approved):

src/gpu_ib/gda_device.cpp
```diff
@@ -281,7 +282,12 @@ GDADevice::GDADevice(MPI_Comm _comm) : comm(_comm), heap(_comm) {
   }
   char* value{nullptr};
   if ((value = getenv("ROCSHMEM_USE_IB_HCA"))) {
-    requested_dev = value;
+    requested_dev = strdup(value);
+  } else {
+    int gpu_dev = 0;
+    CHECK_HIP(hipGetDevice(&gpu_dev));
+    int nic_dev = rocshmem::GetClosestNicToGpu(gpu_dev, &requested_dev);
+    assert (nic_dev != -1);
```
Yiltan, line 290: I think you'd still need to use `nic_dev` somewhere

Maybe around here: https://github.com/ROCm/rocSHMEM/blob/gda_devel/src/gpu_ib/gda_device.cpp#L307

(Assuming `nic_dev` provides the same enumeration ordering as `ibv_get_device_list`)

edgargabriel: no, the code later uses the requested_dev variable, nic_dev is the position of the selected device in the array returned by list_devices, requested_dev is the name of the device, and rocSHMEM is using the latter. They should both point to the same device howeer.

Yiltan (approved):

PR #129, opened 2025-05-15
gpu_ib: Cleanups to Mlx5 provider to ease Ionic integration

andrewb24:
Keep both pd_orig and pd_parent.
Add some helpers for lane mask etc.
Add generic defines in a few places.

### DIFF
--- a/src/gpu_ib/gda_device.cpp
+++ b/src/gpu_ib/gda_device.cpp
@@ -79,7 +79,7 @@ static void dump_ibv_context(struct ibv_context* x) {
    */
   DPRINTF("\n"
          "===============================================\n"
-         "                MLX IBV_CONTEXT\n"
+         "                IBV_CONTEXT\n"
          "===============================================\n"
          "  (ibv_device*)        device              = %p\n"
          "  (int)                cmd_fd              = %d\n"
@@ -103,7 +103,7 @@ static void dump_ibv_device(struct ibv_device* x) {
    */
   DPRINTF("\n"
          "===============================================\n"
-         "               MLX IBV_DEVICE\n"
+         "               IBV_DEVICE\n"
          "===============================================\n"
          "  (enum ibv_node_type)      node_type      = %d\n"
          "  (enum ibv_transport_type) transport_type = %d\n"
@@ -123,7 +123,7 @@ static void dump_ibv_pd(struct ibv_pd* x) {
    */
   DPRINTF("\n"
          "===============================================\n"
-         "               MLX IBV_PD\n"
+         "               IBV_PD\n"
          "===============================================\n"
          "  (ibv_context*) context = %p\n"
          "  (uint32_t)     handle  = 0x%x\n",
@@ -159,7 +159,7 @@ static void dump_ibv_port_attr(struct ibv_port_attr* x) {
    */
   DPRINTF("\n"
          "===============================================\n"
-         "               MLX IBV_PORT_ATTR\n"
+         "               IBV_PORT_ATTR\n"
          "===============================================\n"
          "  (enum ibv_port_state) state           = %u\n"
          "  (enum ibv_mtu)        max_mtu         = %u\n"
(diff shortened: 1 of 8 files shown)


src/gpu_ib/gda_device.cpp
```diff
@@ -599,15 +599,15 @@ void GDADevice::ib_init(struct ibv_device* ib_dev, uint8_t port) {
...
+  GPUIB_CHECK_NNULL(ib_state->pd_orig, "ib allocate pd");
+  dump_ibv_pd(ib_state->pd_orig);
 
-  ibv_parent_domain_init_attr pattr;
+  ibv_parent_domain_init_attr pattr{};
   init_parent_domain_attr(&pattr);
-  ib_state->pd = ibv_alloc_parent_domain(ib_state->context, &pattr);
-  GPUIB_CHECK_NNULL(ib_state->pd, "ibv_alloc_parent_domain");
-  dump_ibv_pd(ib_state->pd);
+  ib_state->pd_parent = ibv_alloc_parent_domain(ib_state->context, &pattr);
```
BKP, line 608: The original PD was ignored in the prior code since we clobbered it with the parent domain.

For Mellanox cards, is it safe to modify the other parts of the code with the orig_pd?

I've never tried to modify this before so I'm not sure what the knock-on effects of the change are.

src/gpu_ib/gda_device.hpp
```diff
@@ -87,7 +88,7 @@ class GDADevice {
       exp_qp_attr.qp_state = IBV_QPS_RTR;
       exp_qp_attr.path_mtu = IBV_MTU_4096;
       exp_qp_attr.ah_attr.sl = 1;
-      exp_qp_attr.max_dest_rd_atomic = 1;
+      exp_qp_attr.max_dest_rd_atomic = GPUIB_MAX_ATOMIC;
```
BKP, line 91: @ahubbe-amd @andrewb24 
What are the consequences of increasing this beyond 1?

andrewb24: I believe it's the number of atomic ops that can be outstanding at any given time.
Note that it will stay at 1 for Mellanox.

src/gpu_ib/gda_device.cpp
andrewb24: @ahubbe-amd Please comment

src/gpu_ib/gda_device.hpp
BKP: I think that it's "the number of responder resources for handling incoming RDMA reads and atomic operations".

I'm wondering about how changing this to a number > 1 would affect transaction ordering.

andrewb24: See IBTA section 9.4.5.3.
```
The responding node saves the reply data, the PSN, and an indication that
the stored data is from an ATOMIC Operation. This saved data is used to
generate the response for retried ATOMIC Operations. Note that the execution
of an RDMA READ operation may consume the same resources as is used to
save the ATOMIC Operation PSN and reply data. The information is stored in
the destination QP’s “connection context”.
```

andrewb24: Everything is still guaranteed to be in-order, atomic, etc. etc.

src/gpu_ib/gda_device.cpp
ahubbe-amd: The parent domain can be used in place of the original pd, but only the parent domain has the allocator override.  The cleanup here is in preparation for ionic gpu_ib, where we will create an additional two parent domains from the original pd.  If rocshmem eventually does its own cleanup and destroying of resources, not just exiting, then it would also be proper to destroy the parent domain(s) and then destroy the original pd.

src/gpu_ib/gda_device.hpp
ahubbe-amd: When the sq is filled with lots of read and atomic requests, these limits are for how many read and atomic requests the nic will send to the remote device before they are acknowledged, and how many remote read and atomic requests this nic can queue up for processing before it has to drop new requests or reply with responder-not-ready.  It's better for this number to be just a bit larger than the read/atomic request rate x round trip response time to avoid stalling the requester.

ahubbe-amd (approved):

BKP (approved):

andrewb24:
@BKP Can I get you to pull this to your system and give it a run? We don't have any Mlx hardware here to test.

BKP:
I validated that the changes are fine for MLX.

andrewb24:
> I validated that the changes are fine for MLX.

Thank you!

PR #130, opened 2025-05-16
add hip cmake path

yiakwy-xpu-ml-framework-team:
## Fix build

The build scirpt does not add cmake hip path correctly in latest build

### DIFF
--- a/scripts/build_configs/ipc_single
+++ b/scripts/build_configs/ipc_single
@@ -33,6 +33,8 @@ else
   install_path=$1
 fi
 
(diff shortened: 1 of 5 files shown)


yiakwy-xpu-ml-framework-team:
@abouteiller could you have a look at it ?

abouteiller:
I would prefer we fix the automatic detection of rocm in our cmake system directly, please provide counsel on #136

abouteiller:
#136 should include a fix for this particular problem.

PR #136, opened 2025-05-20
Rework cmakery:

abouteiller:
* detect rocm/hip/rocshmem better, make sure that ROCM_PATH and ROCM_ROOT don't conflict and are taken by default
  * add /opt/rocm as a fallback when nothing else found
  * obtain hipcc in a sanitized way (ensure we use the same logic we use to later find_package hip)
  * factorize redundancies
  * export GPU_TARGETS as part of the cmake target for librocshmem, this helps with a clean error when an application tries to link with the wrong offload-target flag (rather than a cryptic link error)
  * phased out ROCSHMEM_HOME, in favor of rocshmem_ROOT (the cmake blessed way)
  * fix warning 'deprecated find_package(rocm)'


abouteiller:
This supersedes #130

CMakeLists.txt
```diff
@@ -167,37 +147,32 @@ if (NOT BUILD_TESTS_ONLY)
   #############################################################################
   # LINKING AND INCLUDE DIRECTORIES
   #############################################################################
+  target_compile_options(
+    ${PROJECT_NAME}
+    PUBLIC
+      ${offload_flags}
```
abouteiller, line 159: Putting this in PUBLIC exports it to `find_package(rocshmem)` consumers that use `*_link_libraries(roc::rocshmem)`; this will cause an error if they provide a different set of `--offload-target` that is incompabible (good), this also forces them to compile all variants we compiled in the library (bad), keep?

cmake/setup_project.cmake
```diff
@@ -0,0 +1,42 @@
...
+###############################################################################
+# DEFAULT BUILD TYPE
+###############################################################################
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING
+      "build type: Release, Debug, RelWithDebInfo, MinSizeRel")
+
+###############################################################################
+# GLOBAL COMPILE FLAGS
+###############################################################################
+list(APPEND CMAKE_PREFIX_PATH ${ROCM_ROOT} $ENV{ROCM_ROOT} ${ROCM_PATH} $ENV{ROCM_PATH})
```
abouteiller: Missing `hip_ROOT`; maybe this needs to be a PREPEND?

This append is required for the find_program below, we could reset CMAKE_PREFIX_PATH after, or we could keep it. There is a subtle difference in find order XXX_ROOT > CMAKE_PREFIX_PATH > ENV{PATH} > HINTS. If we keep the CMAKE_PREFIX_PATH we can remove the HINTS in the find_package calls. choice.

examples/CMakeLists.txt
```diff
@@ -21,6 +21,23 @@
...
+
+add_compile_options($<$<CONFIG:Debug>:-ggdb>)
+add_link_options($<$<CONFIG:Debug>:-ggdb>)
+
+set(CMAKE_HIP_ARCHITECTURES OFF)
+
+###############################################################################
+# PROJECT
+###############################################################################
+project(rocshmem_examples VERSION 1.0.0 LANGUAGES HIP CXX)
```
abouteiller: trying this approach with using LANGUAGE HIP, one open question is how much control we have with choosing the hip/rocm version this way.

tests/functional_tests/CMakeLists.txt
```diff
@@ -22,21 +22,27 @@
...
+# PROJECT
+###############################################################################
+include(${CMAKE_SOURCE_DIR}/cmake/setup_project.cmake)
+project(rocshmem_functional_tests VERSION 1.0.0 LANGUAGES CXX)
 
 ###############################################################################
 # SOURCES
 ###############################################################################
-add_executable(${TESTS_NAME} "")
+add_executable(${PROJECT_NAME} "")
```
abouteiller, line 36: this change the name of the executable, that may break CI? Check.

tests/unit_tests/CMakeLists.txt
```diff
@@ -107,65 +78,32 @@ endif()
 ###############################################################################
 # ROCSHMEM DEPENDENCY
 ###############################################################################
-find_package(hip REQUIRED)
+find_package(hip REQUIRED HINTS ${ROCM_PATH} $ENV{ROCM_PATH} PATHS /opt/rocm)
 
 if (BUILD_TESTS_ONLY)
   find_package(MPI REQUIRED)
+  find_package(rocshmem REQUIRED HINTS $ENV{ROCSHMEM_HOME} ${ROCM_PATH} $ENV{ROCM_PATH} PATHS /opt/rocm)
```
abouteiller: We are phasing out ROCSHMEM_HOME, if it has had no significant uses I would straight up remove it in favor of rocshmem_ROOT.

tests/functional_tests/CMakeLists.txt
Yiltan: Yup, this breaks CI.

Fix in CI: https://github.com/ROCm/rocJenkins/commit/cda88e4105bd969c5c5a460b213f6342e711a695

New build with fix: http://math-ci.amd.com/job/main/job/precheckin/job/rocSHMEM/job/PR-136/

CMakeLists.txt
```diff
@@ -36,6 +36,12 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR AND
   message(FATAL_ERROR "In-source build")
 endif()
 
+###############################################################################
+# DEFAULT BUILD TYPE
+###############################################################################
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING
```
Yiltan: Do we still need this line with the change in line 77: `include(${CMAKE_SOURCE_DIR}/cmake/setup_project.cmake)`

CMakeLists.txt
```diff
@@ -135,6 +100,14 @@ if (NOT BUILD_TESTS_ONLY)
   #############################################################################
   # SET GPU ARCHITECTURES
   #############################################################################
+  include(cmake/rocm_local_targets.cmake)
+
+  set(DEFAULT_GPUS
```
Yiltan, line 109: Do we still use this?

CMakeLists.txt
```diff
@@ -121,9 +80,15 @@ else()
 endif()
 message(STATUS "rocSHMEM Version: " "${VERSION_STRING}")
 
-rocm_setup_version(VERSION ${VERSION_STRING})
 project(rocshmem VERSION ${VERSION_STRING} LANGUAGES CXX)
 
+find_package(ROCmCMakeBuildTools PATHS /opt/rocm)
```
Yiltan, line 89: Theres a few instances where we reference `/opt/rocm`. Should we be using `ROCM_PATH`?

Context: on a system that uses Lmod, they may have in non-standard locations e.g. `/opt/rocm-x.y`, `/opt/rocm-x.z`, etc

Yiltan (approved):
Few non-critical comments. It can be merged without addressing

CMakeLists.txt
edgargabriel: I think this would be good to have

CMakeLists.txt
abouteiller: Yes, it's the default value set if user does not set GPU_TARGETS

cmake/setup_project.cmake
```diff
@@ -0,0 +1,46 @@
...
+###############################################################################
+# DEFAULT BUILD TYPE
+###############################################################################
+set(CMAKE_BUILD_TYPE "Release" CACHE STRING
+      "build type: Release, Debug, RelWithDebInfo, MinSizeRel")
+
+###############################################################################
+# GLOBAL COMPILE FLAGS
+###############################################################################
+foreach (root ${hip_ROOT} $ENV{hip_ROOT} ${ROCM_ROOT} $ENV{ROCM_ROOT} ${ROCM_PATH} $ENV{ROCM_PATH})
```
abouteiller, line 34: Place where we capture ROCM_PATH and co

CMakeLists.txt
abouteiller: https://github.com/ROCm/rocSHMEM/pull/136/files#r2150610112

edgargabriel (approved):

PR #138, opened 2025-05-21
Multi-Node rocshmem_finalize() bug

Yiltan:
# Problem

A regression occurred where we cannot run a simple multi-node program like so:

```
int main (int argc, char **argv)
{
    char* ompi_local_rank = getenv("OMPI_COMM_WORLD_LOCAL_RANK");
    CHECK_HIP(hipSetDevice(atoi(ompi_local_rank)));

    rocshmem_init();
    rocshmem_finalize();
    return 0;
}
```

## Source of Problem
`MPI_Init_thread` is used throughout the library and it could take a code path where rocSHMEM is not initialized by the `MPIInitSingleton` object. For example, if it is first initialized by  the remote heap info, the `MPIInitSingleton` would assume that we initialized MPI externally to rocSHMEM and would not call `MPI_Finalize` on exit. Thus create not exiting cleanly.

# Solution
- Only call `MPI_Init_thread` in a single location within the library. 
- Not allow `rocshmem_my_pe()` to be called before `rocSHMEM init`

# Code changes
- Updated examples and functional tests to set the device based on an Open MPI environment variable. 
- Move initialization code into the MPI singleton.
   - This breaks the semantics of a singleton but it is required as we need to support sub communicators.
   - I am open to alternative names to this class.

### DIFF
--- a/examples/rocshmem_allreduce_test.cc
+++ b/examples/rocshmem_allreduce_test.cc
@@ -123,15 +123,13 @@ int main (int argc, char **argv)
         nelem = atoi(argv[1]);
     }
 
-    int my_pe = rocshmem_my_pe();
-    int npes =  rocshmem_n_pes();
-
-    int ndevices, my_device = 0;
-    CHECK_HIP(hipGetDeviceCount(&ndevices));
-    my_device = my_pe % ndevices;
-    CHECK_HIP(hipSetDevice(my_device));
+    char* ompi_local_rank = getenv("OMPI_COMM_WORLD_LOCAL_RANK");
+    CHECK_HIP(hipSetDevice(atoi(ompi_local_rank)));
 
     rocshmem_init();
+
+    int my_pe = rocshmem_my_pe();
+    int npes =  rocshmem_n_pes();
 
     int *source = (int *)rocshmem_malloc(nelem * sizeof(int));
     int *dest = (int *)rocshmem_malloc(nelem * sizeof(int));
--- a/examples/rocshmem_allreduce_test.cc
+++ b/examples/rocshmem_allreduce_test.cc
@@ -54,19 +54,9 @@
 
  */
 
-#include <iostream>
-
-#include <hip/hip_runtime_api.h>
-#include <hip/hip_runtime.h>
 #include <rocshmem/rocshmem.hpp>
 
-#define CHECK_HIP(condition) {                                            \
-        hipError_t error = condition;                                     \
-        if(error != hipSuccess){                                          \
-            fprintf(stderr,"HIP error: %d line: %d\n", error,  __LINE__); \
-            MPI_Abort(MPI_COMM_WORLD, error);                             \
-        }                                                                 \
-    }
+#include "util.h"
 
 using namespace rocshmem;
 
@@ -123,8 +113,7 @@ int main (int argc, char **argv)
         nelem = atoi(argv[1]);
     }
 
-    char* ompi_local_rank = getenv("OMPI_COMM_WORLD_LOCAL_RANK");
-    CHECK_HIP(hipSetDevice(atoi(ompi_local_rank)));
+    CHECK_HIP(hipSetDevice(get_launcher_local_rank()));
 
     rocshmem_init();
 
--- a/examples/util.h
+++ b/examples/util.h
@@ -1,0 +1,52 @@
+/******************************************************************************
+ * Copyright (c) Advanced Micro Devices, Inc. All rights reserved.
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to
+ * deal in the Software without restriction, including without limitation the
+ * rights to use, copy, modify, merge, publish, distribute, sublicense, and/or
+ * sell copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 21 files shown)


examples/rocshmem_allreduce_test.cc
```diff
@@ -123,16 +123,14 @@ int main (int argc, char **argv)
...
     }
 
-    int my_pe = rocshmem_my_pe();
-    int npes =  rocshmem_n_pes();
-
-    int ndevices, my_device = 0;
-    CHECK_HIP(hipGetDeviceCount(&ndevices));
-    my_device = my_pe % ndevices;
-    CHECK_HIP(hipSetDevice(my_device));
+    char* ompi_local_rank = getenv("OMPI_COMM_WORLD_LOCAL_RANK");
```
edgargabriel: I would prefer if we would introduce a function in a new header file in the examples directory (which is included by all files in this directory) that returns the local rank. The initial version of that routine can do the `getenv("OMPI_COMM_WORLD_LOCAL_RANK")`, but we might add additional mechanisms later, e.g. for MPICH/CrayMPICH and SLURM (they have separate environment variables)

abouteiller: That makes the examples Open MPI specific, but I don't have a great alternative to propose

edgargabriel: This is from the Osu benchmarks:
```
int omb_get_local_rank()
{
    char *str = NULL;
    int local_rank = -1;

    if ((str = getenv("MV2_COMM_WORLD_LOCAL_RANK")) != NULL) {
        local_rank = atoi(str);
    } else if ((str = getenv("MVP_COMM_WORLD_LOCAL_RANK")) != NULL) {
        local_rank = atoi(str);
    } else if ((str = getenv("OMPI_COMM_WORLD_LOCAL_RANK")) != NULL) {
        local_rank = atoi(str);
    } else if ((str = getenv("MPI_LOCALRANKID")) != NULL) {
        local_rank = atoi(str);
    } else if ((str = getenv("SLURM_PROCID")) != NULL) {
        local_rank = atoi(str);
    } else if ((str = getenv("LOCAL_RANK")) != NULL) {
        local_rank = atoi(str);
```

Yiltan: These changes have been applied: https://github.com/ROCm/rocSHMEM/pull/138/commits/42b38a772160e7bd3214c5ced0c0ee274921c30c

edgargabriel:
Do we know why the gfx942 tests are failing the way they are right now? Is this something on the new system? Otherwise the PR looks good to me.

Yiltan:
> Do we know why the gfx942 tests are failing the way they are right now? Is this something on the new system? Otherwise the PR looks good to me.

Yup, the gfx942 errors are due to the incorrect configuration of the NICs on the new system. We could use the `--without-verbs` flag in the UCX build to get CI to pass as a workaround but that is just avoidance

edgargabriel:
> > Do we know why the gfx942 tests are failing the way they are right now? Is this something on the new system? Otherwise the PR looks good to me.
> 
> Yup, the gfx942 errors are due to the incorrect configuration of the NICs on the new system. We could use the `--without-verbs` flag in the UCX build to get CI to pass as a workaround but that is just avoidance

in that case we could also just add `-x UCX_TLS=sm,rocm,tcp,self` to the mpirun options

scripts/functional_tests/driver.sh
```diff
@@ -141,6 +141,9 @@ ExecTest() {
   if [[ "" != "$HOSTFILE" ]]
   then
     OPTIONS+=" --hostfile $HOSTFILE"
+  else
+    # Temporary workaround for CI bug
+    OPTIONS+=" -x UCX_TLS=sm,rocm,tcp,self"
```
abouteiller: Not a big fan of this as this changes behavior on all platforms and in interactive uses. Can't we have that in the CI config instead since it is specific to these systems?

examples/util.h
```diff
@@ -0,0 +1,52 @@
...
+        if(error != hipSuccess){                                          \
+            fprintf(stderr,"HIP error: %d line: %d\n", error,  __LINE__); \
+            MPI_Abort(MPI_COMM_WORLD, error);                             \
+        }                                                                 \
+    }
+
+static int get_launcher_local_rank() {
+    char *local_rank_str = nullptr;
+
+    local_rank_str = getenv("OMPI_COMM_WORLD_LOCAL_RANK");
```
abouteiller, line 44: I thought you had imported
https://github.com/ROCm/rocSHMEM/pull/138#discussion_r2100491548

src/mpi_init_singleton.cpp
```diff
@@ -26,18 +26,20 @@
 
 namespace rocshmem {
 
-MPIInitSingleton* MPIInitSingleton::instance{nullptr};
-
-MPIInitSingleton::MPIInitSingleton() {
+MPIInitSingleton::MPIInitSingleton(MPI_Comm comm) {
```
abouteiller: The name is now misleading, what about `MPIInstance::`

scripts/functional_tests/driver.sh
Yiltan: I agree, I will actually remove this. It seems to break atomics in UCX

edgargabriel: yeah, this was meant as a temporary work around, would have the same impact as compiling UCX with `--without-verbs`

src/mpi_init_singleton.cpp
Yiltan: Renamed

edgargabriel (approved):
LGTM

examples/util.h
Yiltan: I just imported the ones that I have tested. 

Unfortunately, I haven't had time to test mvapich/mpich/slurm/etc. This should easily be added in the future

abouteiller: Sounds good, lets have a tracking issue to do this later

abouteiller (approved):

examples/util.h
Yiltan: `LWPRHMEM-135` will be tracking this issue

PR #150, opened 2025-06-04
added init example and all_reduce example on the files

akolliasAMD:
Code coverage script updates

### DIFF
--- a/scripts/build_configs/codecov
+++ b/scripts/build_configs/codecov
@@ -29,6 +29,11 @@ LLVM_PROFILE_FILE="$PROFRAW_DIR/ro_net-unit-%p.profraw" ../scripts/unit_tests/dr
 
 LLVM_PROFILE_FILE="$PROFRAW_DIR/ro_ipc-unit-%p.profraw" ../scripts/unit_tests/driver.sh ./ro_ipc/tests/unit_tests/rocshmem_unit_tests all
 
+# Examples coverage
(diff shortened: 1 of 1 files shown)


abouteiller (dismissed):

scripts/build_configs/codecov
```diff
@@ -29,6 +29,11 @@ LLVM_PROFILE_FILE="$PROFRAW_DIR/ro_net-unit-%p.profraw" ../scripts/unit_tests/dr
 
 LLVM_PROFILE_FILE="$PROFRAW_DIR/ro_ipc-unit-%p.profraw" ../scripts/unit_tests/driver.sh ./ro_ipc/tests/unit_tests/rocshmem_unit_tests all
 
+# Examples coverage
+LLVM_PROFILE_FILE="$PROFRAW_DIR/ipc_unit-%m.profraw"  mpirun -np 4 ./ipc/examples/rocshmem_init_attr_test
+LLVM_PROFILE_FILE="$PROFRAW_DIR/ipc_unit-%m.profraw"  mpirun -np 4 ./ipc/examples/rocshmem_allreduce_test
```
Yiltan: I assume we only need this for IPC and not RO or RO+IPC?

edgargabriel: for the init_attr_test that is definitely correct, there is no additional benefit from running at with all different conduits. For the allreduce_test, I am honestly a bit surprised on why that increases the coverage, since allreduce is already also tested as part of the functional tests, isn't it?

edgargabriel (dismissed):

edgargabriel (approved):

abouteiller (approved):

PR #153, opened 2025-06-09
Introduce support for executing the IPC conduit without MPI

edgargabriel:
This PR adds the capability to execute rocSHMEM jobs using the IPC conduit without necessitating to initialize the MPI library. MPI is still required to compile rocSHMEM with the IPC conduit. However, depending on the initialization method MPI might not be used. Specifically:

- initializing using `rocshmem_init` or `rocshmem_init_threads` still will use MPI internally
- initializing using `rocshmem_init_attr` with `ROCSHMEM_INIT_WITH_MPI_COMM` will still use MPI internally
- initializing using `rocshmem_init_attr` with `ROCSHMEM_INIT_WITH_UNIQUEID` will *not* use MPI by default. However, the user can force using MPI even in this scenario by setting `ROCSHMEM_UNIQUEID_WITH_MPI=1`. 

The unit tests are using the MPI path only. The functional tests have been adjusted to be able to run the non-MPI path as well. Specifically, running the non-MPI path in the functional tests requires detecting the PMIX libraries, and setting the environment variable `ROCSHMEM_TEST_UUID=1`.

Known limitations:

- at the moment, `rocshmem_team_split_strided` only works for the non-MPI path if the parent team contains *all* processes that were used within the same instance of `rocshmem_init_attr`. 
- the unit tests are not testing some of the new classes and abstractions introduced. This might be necessary going forward, but it will be a non-trivial endeavor since the unit-tests are initializing MPI first thing in the gtest main function.
- there is probably room for improvement from the code cleanliness perspective. The separation of the MPI and non-MPI path are achieved through separate classes in some instances (e.g. `CommunicatorMPI` vs. `CommunicatorTCP` in `RemoteHeapInfo`), or by using `comm=MPI_COMM_NULL` or `bootstrap=nullptr` in some other instances. Further iterations to properly clean-up the handling are probably required, but this code is functional as is.
- The PMIX detection probably will need to be adjusted to work in the most generic setting, but it works as is on my platform.


edgargabriel:
As an interesting little tid-bit, running the functional tests using the MPI initialization path takes on the MI250X node 7m43s, while with the non-MPI path it takes 5m31s, so noticeably faster.

edgargabriel:
argh, this is what I was afraid of with the PMIx detection. Some of the builds are failing becase we find different pmix header file and libraries:

```

-- Found pmix.h at /var/jenkins_home/workspace/main_precheckin_rocSHMEM_PR-153/ompi/install/include
-- Found libpmix at /usr/lib/x86_64-linux-gnu/libpmix.so

```

CMakeLists.txt
```diff
@@ -174,6 +173,11 @@ if (NOT BUILD_TESTS_ONLY)
   set(THREADS_PREFER_PTHREAD_FLAG TRUE)
   find_package(Threads REQUIRED)
 
+  include(cmake/find_pmix.cmake)
+  check_pmix()
+
+configure_file(cmake/rocshmem_config.h.in rocshmem_config.h)
```
Yiltan: Picky comment but indent:

```suggestion
  configure_file(cmake/rocshmem_config.h.in rocshmem_config.h)
```

src/memory/window_info.hpp
```diff
@@ -180,17 +246,9 @@ class WindowInfo {
...
-   */
-  void* win_start_{nullptr};
-
-  /**
-   * @brief Raw pointer marking the end of window
-   */
-  void* win_end_{nullptr};
 };
 
+
```
Yiltan:
```suggestion
```

cmake/find_pmix.cmake
```diff
@@ -0,0 +1,58 @@
...
+# IN THE SOFTWARE.
+###############################################################################
+
+# Find pmix installation.
+# Two different scenarios need to be covered:
+#  - pmix installed as part of Open MPI, i.e. it will be in the MPI installation directories
+#  - pmix deployed with linux distros
+#  - later: handle pmix deployed with slurm.
+
+macro(check_pmix)
```
Yiltan, line 31: I seem to be getting the warning: 

```
CMake Warning at tests/functional_tests/CMakeLists.txt:30 (add_executable):
  Cannot generate a safe runtime search path for target
  rocshmem_example_driver because there is a cycle in the constraint graph:

    dir 0 is [/usr/local/lib]
      dir 1 must precede it due to runtime library [libmpi.so.40]
    dir 1 is [/home/ytemucin/install/ompi/lib]
      dir 0 must precede it due to runtime library [libpmix.so.2]

  Some of these libraries may not be found correctly.
  ```
  
  I've been able to run programs fine.

File: src/util.cpp [hunk 4]
```diff
@@ -137,6 +137,11 @@ rocshmem_env_config::rocshmem_env_config() {
     ro_progress_delay = atoi(env_value);
   }
 
+  env_value = getenv("ROCSHMEM_UNIQUEID_NO_MPI");
```
Yiltan: Could add here: 

https://github.com/ROCm/rocSHMEM/blob/develop/docs/compile_and_run.rst?plain=1#L81

tests/functional_tests/test_driver.cpp
```diff
@@ -28,6 +28,104 @@
...
+        abort();
+      }
+    }
+
+    /* call fence to synchronize with our peers - instruct
+     * the fence operation to collect and return all "put"
+     * data from our peers */
+    PMIX_INFO_CREATE(info, 1);
+    flag = true;
+        PMIX_INFO_LOAD(info, PMIX_COLLECT_DATA, &flag, PMIX_BOOL);
```
Yiltan, line 97: PMIX_INFO_LOAD == fence operation?

Yiltan (dismissed):

cmake/find_pmix.cmake
edgargabriel: Hm, I thought I fixed this particular issue. However, it is very clear that the cmake detection of pmix is incomplete and will require some more work. My hope was that we can merge the PR to allow for future work to continue, and we can improve the cmake logic in a later commit.

tests/functional_tests/test_driver.cpp
edgargabriel: no, the PMIx_Fence() is in the next line. I just followed in this code sequence an example from the PMIx repository.

cmake/find_pmix.cmake
Yiltan: I agree, the cmake issue can be resolved later

src/util.cpp
edgargabriel: done. I noticed that some other environment variables are also missing there (e.g. RO_DISABLE_IPC, or the Bootstrap envs), I will add them independent of this PR.

Yiltan (dismissed):

docs/compile_and_run.rst
```diff
@@ -81,3 +81,6 @@ You can control the behavior of rocSHMEM by using the following environment vari
     * - ROCSHMEM_MAX_NUM_TEAMS
       - 40
       - Defines the number of teams an application can use.
+    * - ROCSHMEM_UNIQUEID_NO_MPI
```
abouteiller: ROCSHMEM_UNIQUEID_WITH_MPI

src/host/host.cpp
```diff
@@ -144,6 +149,41 @@ __host__ void HostInterface::create_hdp_window() {
...
+  size_t pool_size = max_num_ctxs_ * sizeof(HostContextWindowInfo*);
+  host_window_context_pool_ =
+      reinterpret_cast<HostContextWindowInfo**>(malloc(pool_size));
+
+  for (int ctx_i = 0; ctx_i < max_num_ctxs_; ctx_i++) {
+    host_window_context_pool_[ctx_i] =
+        new HostContextWindowInfo(heap);
+  }
+
+#if !defined(USE_COHERENT_HEAP) && !defined(USE_SINGLE_NODE)
```
abouteiller: this will conflict with develop, presumably a merge conflict, should be 

```
-#if !defined(USE_COHERENT_HEAP) && !defined(USE_SINGLE_NODE)
+#if defined USE_HDP_FLUSH && not defined USE_SINGLE_NODE
```

tests/functional_tests/test_driver.cpp
abouteiller: The info controls the behavior of the fence here, looks correct.

src/host/host.cpp
edgargabriel: done, thanks for identifying this!

Yiltan (dismissed):

src/ipc/backend_ipc.cpp
```diff
@@ -315,8 +390,13 @@ void IPCBackend::init_wrk_sync_buffer() {
   /*
    * all-to-all exchange with each PE to share the IPC handles.
    */
-  MPI_Allgather(MPI_IN_PLACE, sizeof(hipIpcMemHandle_t), MPI_CHAR,
-                ipc_handle, sizeof(hipIpcMemHandle_t), MPI_CHAR, backend_comm);
+  if (backend_comm != MPI_COMM_NULL) {
+    MPI_Allgather(MPI_IN_PLACE, sizeof(hipIpcMemHandle_t), MPI_CHAR,
```
abouteiller, line 394: We could hide the MPI_Allgather in backend_bootstr->allgather, but that is a future improvement not needed now.

src/rocshmem.cpp
```diff
@@ -122,57 +199,17 @@ rocshmem_ctx_t ROCSHMEM_HOST_CTX_DEFAULT;
...
-
-      bootstr.initialize(attr->uid, rocshmem_env_.get_bootstrap_timeout());
-      int *inc_ranks = new int[attr->nranks];
-      inc_ranks[attr->rank] = world_rank;
+    assert (attr->nranks > 0);
+    assert (attr->rank >= 0);
+    assert (attr->rank < attr->nranks);
 
-      bootstr.allGather (inc_ranks, sizeof(int));
+    bootstr = new TcpBootstrap(attr->rank, attr->nranks);
```
abouteiller, line 206: we always create a bootstrap even if init_with_mpi?

tests/unit_tests/symmetric_heap_gtest.hpp
```diff
@@ -37,7 +37,7 @@ class SymmetricHeapTestFixture : public ::testing::Test
     /**
      * @brief Symmetric heap object
      */
-    SymmetricHeap symmetric_heap_ {};
+    SymmetricHeap symmetric_heap_ {MPI_COMM_WORLD};
```
abouteiller, line 40: We don't test the symmetric heap without MPI, that may need added later

abouteiller (approved):

src/rocshmem.cpp
edgargabriel: yes, in the UniqueID initialization part we do still need the bootstrap object.

src/ipc/backend_ipc.cpp
edgargabriel: I think the proper solution is to have method that takes a communicator as an argument (in which case we do MPI_Allgather, and a method that takes the bootstrap as an argument, in which case we do the bootstrap->allGather(). I would not add MPI code into bootstrap class.

PR #158, opened 2025-06-13
Swdev/536571 with additional issues found for other various missing includes

abouteiller:
* device side assert needs including `<hip/hip_runtime.h>` in addition to `<cassert>`
* missing include `<thread>`
* replace some assert(false) with abort in cases where -DNDEBUG would cause code to silently produce wrong results
* disable IPC testers for not-implemented AMOs in that transport.

### DIFF
--- a/src/ipc/context_ipc_tmpl_device.hpp
+++ b/src/ipc/context_ipc_tmpl_device.hpp
@@ -32,6 +32,8 @@
 #include "ipc_team.hpp"
 #include "../rocshmem_calc.hpp"
 
+#include <hip/hip_runtime.h>
+
 namespace rocshmem {
 
 /******************************************************************************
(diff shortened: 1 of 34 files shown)


src/ipc/context_ipc_tmpl_device.hpp
```diff
@@ -92,41 +94,41 @@ __device__ void IPCContext::amo_set(void *dest, T value, int pe) {
...
 __device__ T IPCContext::amo_fetch_and(void *dst, T value, int pe) {
-  assert(false);
+  abort();
   return 0;
 }
 
 template <typename T>
 __device__ void IPCContext::amo_and(void *dst, T value, int pe) {
-  assert(false);
+  abort();
```
Yiltan, line 112: Nice catch! 

I wonder why this was passing in CI prior to this PR, I'm guessing there could be an issue with the functional test.

After looking, The default buffer values are the same as the expected values

edgargabriel:
well, this is a major milestone, *all* tests have passed, including the gfx942 ones!

edgargabriel (approved):

PR #163, opened 2025-06-18
use uncached_load instructions for the test()

edgargabriel:
this is also already done on develop, so this PR just makes the gda_devel branch use the same mechanism.


edgargabriel:
Closing for now, we can decide later whether this is the right approach or not.

PR #165, opened 2025-06-20
Add amo fetch add and non_fetch add for self (same-rank) atomic updates

BKP:


edgargabriel:
@BKP this PR doesn't look right. Do you want to file the PR against gda_devel branch maybe?

abouteiller (approved):

edgargabriel:
Is the change in queue_pair.cpp something that we definitely need? If yes, we should go ahead and merge the PR to avoid segmentation and reduce the nubmer of additional patches that we need to keep track of.

avinashkethineedi (approved):

avinashkethineedi:
> Is the change in queue_pair.cpp something that we definitely need? If yes, we should go ahead and merge the PR to avoid segmentation and reduce the nubmer of additional patches that we need to keep track of.

yes, it is definitely needed for the fetching atomics

abouteiller:
All features have been duplicated in merged #167

PR #174, opened 2025-06-25
Prevent send queue counter ordering bug

BKP:


edgargabriel:
@BKP would it be possible to rebase your PR on top of what is already in gda_devel, such that we only get the actual diff? Your base branch that you applied the modifications on seems to be something else.

BKP:
Closed duplicate

PR #178, opened 2025-06-26
Check for counter load order update in send queue

BKP:


edgargabriel (approved):

abouteiller (approved):
passed on hour of deepep torture test

PR #179, opened 2025-06-26
non-fetching amos can be implicit  nbi

abouteiller:
so there is no need to terminal quiet

this passed 1h of deepep torture test. Performance eval pending.


abouteiller:
correctness with deepEP has been verified

edgargabriel (approved):

PR #182, opened 2025-06-27
fixed compilation targets on cmake

akolliasAMD:


edgargabriel (dismissed):

CMakeLists.txt
```diff
@@ -122,23 +122,20 @@ if (NOT BUILD_TESTS_ONLY)
...
-    set(SUPPORTED_GPUS ${GPU_TARGETS})
+    set(SUPPORTED_GPUS ${CURRENT_GPU_TARGETS})
   endif()
 
-  set(COMPILING_TARGETS "${SUPPORTED_GPUS}" CACHE STRING "GPU targets to compile for.")
-  message(STATUS "Compiling for ${COMPILING_TARGETS}")
+  set(GPU_TARGETS "${SUPPORTED_GPUS}" CACHE STRING "GPU architectures to compile for")
 
-  foreach (target ${COMPILING_TARGETS})
-    list(APPEND offload_flags --offload-arch=${target})
```
abouteiller, line 140: This removes the inclusion of the offload-target from the rocshmem::rocshmem cmake target, see (https://github.com/ROCm/rocSHMEM/pull/182/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR159, were this is set public, and hence inherited by consumers) that will make external consumers of the library not abide by the offload targets we compiled in. In particular that breaks the BUILD_TEST_ONLY option.

abouteiller (approved):

PR #185, opened 2025-06-30
Let tests compile when pmix is not found

abouteiller:
there was a typo, we would still try to compile with pmix when pmix is not found./


edgargabriel:
@abouteiller  are we setting HAVE_PMIX somewhere in this modified version? The test_driver.cpp uses the HAVE_PMIX macro internally

abouteiller:
> @abouteiller are we setting HAVE_PMIX somewhere in this modified version? The test_driver.cpp uses the HAVE_PMIX macro internally

yes, on modified line 93

We are using PMIx_FOUND in cmake for the case where we did find_package(pmix) and we don't have to do it ourselves, we still use HAVE_PMIX in the C++ code.

edgargabriel (approved):

abouteiller:
This change not needed in mainline for 7.0, only for devel

PR #186, opened 2025-06-30
rocshmem_config.h has a different include path when installed and built-dir

abouteiller:
Having rocshmem_config.h with different include path in build and install cases causes functional tests and unit tests to fail building in BUILD_TEST_ONLY configs.


abouteiller:
This change will need to make its way to mainline before 7.0

akolliasAMD (approved):
LGTM

PR #188, opened 2025-07-02
Make gda_devel branch work without MPI library 

edgargabriel:
This PR integrates the no-MPI execution path from the develop branch / IPC conduit to the gda_devel branch.

Some comments:
 - Similarly to the work on the IPC conduit, an MPI library is still required to compile rocSHMEM. However, if an application uses the unique_id initiatlization method, rocSHMEM can execute without MPI support.
 - The PMIx detection for the functional tests is not setting the flag on this branch correctly, hence the user/tester will  have to add for now manually an `#define HAVE_PMIX 1` into the functional_tests/test_driver.cpp file to compile with PMIx support. Fixing this would have required to bring the entire CMake changes from develop to this branch, which would have made the actual changes in the PR more difficult to identify.
 - As far as I can tell, all tests that used to pass with MPI are also passing without MPI
 - More testing with deepEp will be required before merging. Note, that for the MPI Path the way rocSHMEM was used so far, we will have to set now the `ROCSHMEM_UNIQUEID_WITH_MPI=1` environment variable to pass.
 - Testing with deepEp without MPI should also be done, but this will require  new instructions for how to compile pytorch.
 - I disabled one test in the unit tests, since the singleton class has been renamed in develop, and I think its pointless to fix the test in this branch. (I need however the modified functionality of the class from develop).

### DIFF
--- a/src/gpu_ib/gda_device.cpp
+++ b/src/gpu_ib/gda_device.cpp
@@ -267,7 +267,30 @@ void dump_mlx5dv_cq(struct mlx5dv_cq *cq_dv, int conn_num) {
 }
 #endif // !GPUIB_IONIC
 
-GDADevice::GDADevice(MPI_Comm _comm) : comm(_comm), heap(_comm) {
+GDADevice::GDADevice(TcpBootstrap* bootstrap):  heap(MPI_COMM_NULL, bootstrap) {
+  init_part1();
+  backend_bootstr = bootstrap;
+
+  my_pe = bootstrap->getRank();
+  num_pes = bootstrap->getNranks();
+
+  host_interface = new HostInterface(bootstrap, &heap);
+  init_part2();
+}
+
+GDADevice::GDADevice(MPI_Comm _comm) : comm(_comm), heap(_comm, nullptr) {
+  init_part1();
+
+  init_mpi_once(_comm);
+  comm = _comm;
+  NET_CHECK(MPI_Comm_size(comm, &num_pes));
+  NET_CHECK(MPI_Comm_rank(comm, &my_pe));
+
+  host_interface = new HostInterface(comm, &heap);
+  init_part2();
+}
+
+void GDADevice::init_part1() {
   CHECK_HIP(hipMalloc(&print_lock, sizeof(*print_lock)));
   *print_lock = 0;
   int* print_lock_addr{nullptr};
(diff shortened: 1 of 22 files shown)


src/gpu_ib/gda_device.cpp
```diff
@@ -414,11 +436,92 @@ __device__ void GDADevice::destroy_ctx(rocshmem_ctx_t *ctx) {
...
+  int num_pes = team_obj->num_pes;
+  int my_pe = team_obj->my_pe;
+  int *pes_in_world = new int[num_pes];
+
+  int my_pe_in_world = team_obj->my_pe_in_world;
+  for (int i = 0; i < num_pes; i++) {
+      pes_in_world[i] = team_obj->get_pe_in_world(i);      
+  }
+
+  // Since this is an in-place algorith, allocate the temporary receive buffer first
```
abouteiller:
```suggestion
  // Since this is an in-place algorithm, allocate the temporary receive buffer first
```

src/gpu_ib/gda_device.cpp
```diff
@@ -414,11 +436,92 @@ __device__ void GDADevice::destroy_ctx(rocshmem_ctx_t *ctx) {
...
+  delete[] recv_buf;
+  delete[] pes_in_world;  
+}
+
+void GDADevice::Allreduce_char_BAND (char* inbuf, char *outbuf, size_t num_bytes,
+	                             Team *team) {
+
+  // Implement an Allreduce outside of MPI. This is specialized for the scenario
+  // required for the team creation, i.e. assuming bytes and using BAND operation.
+  // Implementation uses an Allgather operation followed a local reduction.
```
abouteiller, line 490: potential optimization: do local_reduction(count=num_bytes), allgather(count=num_pes), local_reduction(count=num_pes), not urgent, presumably should do this in develop eventually

src/host/host_templates.hpp
```diff
@@ -99,7 +103,7 @@ inline MPI_Datatype HostInterface::get_mpi_type() {
 
 #define GET_MPI_TYPE(T, MPI_T)                                    \
   template <>                                                     \
-  inline MPI_Datatype HostInterface::get_mpi_type<T>() {          \
+  inline MPI_Datatype HostInterface::get_mpi_type<T>() { \
```
abouteiller, line 106: no change needed
(feel free to ignore if that puts it more inline with develop)

src/host/host_templates.hpp
```diff
@@ -233,7 +254,7 @@ inline size_t status_entry(size_t nelems, const int *status) {
 
 template <typename T>
 size_t HostInterface::wait_until_any(T* ivars, size_t nelems, const int *status, int cmp, T val, WindowInfo* window_info) {
-  if (!nelems) { return SIZE_MAX; }
+  if (!nelems) { return SIZE_MAX;  }
```
abouteiller, line 257: no change needed

src/util.hpp
```diff
@@ -130,7 +132,7 @@ __device__ __forceinline__ int get_flat_id() {
  * Returns true if the caller's thread flad_id is 0 in its wave.
  */
 __device__ __forceinline__ bool is_thread_zero_in_wave() {
-  return (get_flat_block_id() % __AMDGCN_WAVEFRONT_SIZE) == 0;
```
abouteiller, line 133: Isn't the existing way better?

tests/functional_tests/CMakeLists.txt
```diff
@@ -80,6 +86,7 @@ if (BUILD_TESTS_ONLY)
     ${TESTS_NAME}
     PRIVATE
       ${flags_str}
+      $<$<BOOL:HAVE_PMIX>:HAVE_PMIX=1>
```
abouteiller:
```suggestion
      $<$<BOOL:${HAVE_PMIX}>:HAVE_PMIX=1>      
```

src/util.hpp
edgargabriel: We are not supposed to use `__AMDGCN_WAVEFRONT_SIZE` anymore in the code, that's why it was also removed in develop.

abouteiller (approved):

PR #197, opened 2025-07-09
Implement `rocshmem_ptr` in IPC conduit

avinashkethineedi:
Implements `rocshmem_ptr` API in IPC conduit

### DIFF
--- a/src/ipc/context_ipc_device.cpp
+++ b/src/ipc/context_ipc_device.cpp
@@ -106,6 +106,10 @@ __device__ void IPCContext::quiet() {
 
 __device__ void *IPCContext::shmem_ptr(const void *dest, int pe) {
   void *ret = nullptr;
+  void *dst = const_cast<void *>(dest);
+    uint64_t L_offset =
+        reinterpret_cast<char *>(dst) - ipcImpl_.ipc_bases[my_pe];
+    ret = ipcImpl_.ipc_bases[pe] + L_offset;
   return ret;
 }
 
--- a/tests/functional_tests/shmem_ptr_tester.cpp
+++ b/tests/functional_tests/shmem_ptr_tester.cpp
@@ -98,10 +98,10 @@ __global__ void ShmemPtrTest(int loop, int skip, long long int *start_time,
   }
 
   // For data validation in remote PE
-  if(*available == 1 && get_flat_id() == 0) {
-    local_addr = dest + get_flat_grid_size();
-    int *_available = (int*)rocshmem_ptr((void *)local_addr, 1);
-    *_available = 1;
+  if( get_flat_id() == 0 ) {
+    int *store_avail = (int*)(dest + get_flat_grid_size());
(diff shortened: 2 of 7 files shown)


drprajap (dismissed):

src/ipc/context_ipc_device.cpp
```diff
@@ -106,6 +106,10 @@ __device__ void IPCContext::quiet() {
 
 __device__ void *IPCContext::shmem_ptr(const void *dest, int pe) {
   void *ret = nullptr;
+  void *dst = const_cast<void *>(dest);
+    uint64_t L_offset =
+        reinterpret_cast<char *>(dst) - ipcImpl_.ipc_bases[ipcImpl_.shm_rank];
```
drprajap: one thing for my understanding, in RO backend, it checks for ipc availability with incoming pe, is it for safer operation? and we don't need this here?

avinashkethineedi: We can omit that check in the IPC path because all PEs are guaranteed to reside on the same node, which is validated during [library initialization](https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/backend_ipc.cpp#L74). However, in the RO backend, some ranks may be local while others are remote, so we need to check whether the target PE is local before attempting to retrieve its address.

drprajap: ipcImpl_.shm_rank is not set in IPCContext::IPCContext initialization, so it will give false results, so either can use my_pe like other methods here or set  shm_rank explicitly.

avinashkethineedi: yes, will change it to `my_pe`

akolliasAMD (dismissed):

abouteiller:
I saw some defects in manual testing, lets hold back for now

abouteiller:
Only real issue is crash failing in unsupported cases (e.g., ro-only, or 2-nodes deploy), on-node IPC cases work as intended.

tests/functional_tests/shmem_ptr_tester.cpp
```diff
@@ -31,59 +31,146 @@ using namespace rocshmem;
...
+  for (int i = num_wfs / 2; i > 0; i >>= 1 ) {
+    if(t_id < i) {
+      wf_start_time[t_id] = min(wf_start_time[t_id], wf_start_time[t_id + i]);
+    }
+  }
+
+  // For data validation in remote PE
+  if(*available == 1 && get_flat_id() == 0) {
+    local_addr = dest + get_flat_grid_size();
+    int *_available = (int*)rocshmem_ptr((void *)local_addr, 1);
```
abouteiller:
```suggestion
    shmem_put(local_addr, 1, 1)
```
this will prevent the test from failing silent if the visibility updates from rocshmem_ptr are not enforced

abouteiller (approved):

PR #199, opened 2025-07-11
Only issue a single completion per wavefront

BKP:
Issuing more than one completion per wavefront is wasteful.


abouteiller:
correctness with deepEP has been verified

abouteiller (approved):

PR #200, opened 2025-07-11
Add host APIs for querying device ctx and remote heap pointer

drprajap:
* Host API to query device pointer for ROCSHMEM_DEFAULT_CONTEXT, this is needed to support dynamic module initialization via device kernel library bitcode.
* Host API to query remote symmetric heap pointer that can be used in custom device kernel for RMA operations or issue device to device hip memcpy operation.

### DIFF
--- a/src/context_host.cpp
+++ b/src/context_host.cpp
@@ -32,20 +32,9 @@ __host__ Context::Context(Backend* handle, bool shareable)
     : num_pes(handle->getNumPEs()),
       my_pe(handle->getMyPE()),
       fence_(shareable) {
-
-  char** ipc_bases = new char*[handle->ipcImpl.shm_size];
-
-  CHECK_HIP(hipMemcpy(ipc_bases,
-                  handle->ipcImpl.ipc_bases,
-                  handle->ipcImpl.shm_size * sizeof(char *),
-                  hipMemcpyDeviceToHost));
-
-  ipcImpl_.ipc_bases = ipc_bases;
 }
 
 __host__ Context::~Context() {
-
-  delete[] ipcImpl_.ipc_bases;
 }
 
 /******************************************************************************
--- a/src/context_host.cpp
+++ b/src/context_host.cpp
@@ -31,7 +31,22 @@ namespace rocshmem {
 __host__ Context::Context(Backend* handle, bool shareable)
     : num_pes(handle->getNumPEs()),
       my_pe(handle->getMyPE()),
-      fence_(shareable) {}
+      fence_(shareable) {
+
+  char** ipc_bases = new char*[handle->ipcImpl.shm_size];
+
+  CHECK_HIP(hipMemcpy(ipc_bases,
+                  handle->ipcImpl.ipc_bases,
+                  handle->ipcImpl.shm_size * sizeof(char *),
+                  hipMemcpyDeviceToHost));
+
+  ipcImpl_.ipc_bases = ipc_bases;
+}
+
+__host__ Context::~Context() {
+
+  delete[] ipcImpl_.ipc_bases;
+}
 
 /******************************************************************************
  ********************** CONTEXT DISPATCH IMPLEMENTATIONS **********************
@@ -91,6 +106,12 @@ __host__ void Context::quiet() {
   ctxHostStats.incStat(NUM_HOST_QUIET);
 
   HOST_DISPATCH(quiet());
+}
+
+__host__ void* Context::shmem_ptr(const void* dest, int pe) {
+  ctxHostStats.incStat(NUM_HOST_SHMEM_PTR);
+
+  HOST_DISPATCH_RET_PTR(shmem_ptr(dest, pe));
 }
 
 __host__ void Context::sync_all() {
--- a/src/rocshmem_gpu.cpp
+++ b/src/rocshmem_gpu.cpp
@@ -107,27 +107,6 @@ __host__ void * rocshmem_get_device_ctx() {
 
 }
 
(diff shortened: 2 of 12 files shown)


src/rocshmem_gpu.cpp
```diff
@@ -92,6 +92,44 @@ __device__ void rocshmem_query_thread(int *provided) {
...
+
+__host__ void *rocshmem_ptr(void * dest, int pe){
+  void * ctx = nullptr;
+  void *ret = nullptr;
+  int my_pe = rocshmem_my_pe();
+
+  CHECK_HIP(hipMemcpyFromSymbol(&ctx, HIP_SYMBOL(ROCSHMEM_CTX_DEFAULT),
+                             sizeof(rocshmem_ctx_t)));
+
+  IPCContext * IpcCtx = reinterpret_cast<IPCContext *>(ctx);
```
avinashkethineedi: Use [context](https://github.com/ROCm/rocSHMEM/blob/develop/src/context.hpp#L56) base class instead of [IPCcontext ](https://github.com/ROCm/rocSHMEM/blob/develop/src/ipc/context_ipc_device.hpp#L34) to ensure this API is compatible with all backends.

src/rocshmem_gpu.cpp
```diff
@@ -92,6 +92,44 @@ __device__ void rocshmem_query_thread(int *provided) {
...
+
+  char **host_ipc_base  =(char ** ) malloc(IpcCtx->ipcImpl_.shm_size * sizeof(char **));
```
avinashkethineedi: Memory leak

src/rocshmem_gpu.cpp
```diff
@@ -92,6 +92,44 @@ __device__ void rocshmem_query_thread(int *provided) {
...
+  return ctx;
+
+}
+
+__host__ void *rocshmem_ptr(void * dest, int pe){
+  void * ctx = nullptr;
+  void *ret = nullptr;
+  int my_pe = rocshmem_my_pe();
+
+  CHECK_HIP(hipMemcpyFromSymbol(&ctx, HIP_SYMBOL(ROCSHMEM_CTX_DEFAULT),
```
avinashkethineedi: Ideally, this API should use the default host context to retrieve the relevant heap pointer information.

- [RO host context](https://github.com/ROCm/rocSHMEM/blob/develop/src/reverse_offload/backend_ro.cpp#L110)

drprajap: yes, I came across this RO implementation, but I met an issue in extracting relevant detail from that so thought it might not be setup properly for IPC, let me retry and see what exact issue I see.

avinashkethineedi: [ipcImpl_ ](https://github.com/ROCm/rocSHMEM/blob/develop/src/context.hpp#L502) is only initialized in the device context constructor but not in the host context constructor and also, only the RO backend creates a default host context, while the IPC backend does not. This would have caused the issue in extracting relevant information in the IPC backend.

drprajap: I think it does get initialized in host IPCContext [constructor ](https://github.com/ROCm/rocSHMEM/blob/42e28835adb04bc6712c055951d45a5e240772c1/src/ipc/context_ipc_device.cpp#L44 , I am able to use that, check that out in my next commit and let me know if it makes sense, tested it functionally as well

drprajap:
@avinashkethineedi  any other comments? can we take the changes we discussed offline in a separate PR, this will unblock some of the changes from our side.

avinashkethineedi:
> @avinashkethineedi any other comments? can we take the changes we discussed offline in a separate PR, this will unblock some of the changes from our side.

The API implementation needs to be delegated to the host context class. I’ve added a commit with the necessary changes, please test them before merging.

src/context_host.cpp
```diff
@@ -31,7 +31,22 @@ namespace rocshmem {
...
+                  handle->ipcImpl.ipc_bases,
+                  handle->ipcImpl.shm_size * sizeof(char *),
+                  hipMemcpyDeviceToHost));
+
+  ipcImpl_.ipc_bases = ipc_bases;
+}
+
+__host__ Context::~Context() {
+
+  delete[] ipcImpl_.ipc_bases;
```
drprajap: This is not right, this field carries device side pointer from ipcHostInit() and this will try to free up that instead of host side memory. The init for host side ipc_bases and cleanup needs to move to IpcHostContext class as Context class is used to hold device side allocations as well for device side initialization APIs
the runtime error - free(): invalid pointer is coming from here because of above reason.

avinashkethineedi: there is a `IpcImpl` object in multiple classes, but the `ipcHostInit` is only invoked in the backend class, where it populates the `ipc_bases` with the devices pointers (IPC memory). This information needs to be passed to `IpcImpl` object within the host context class for the host APIs to work correctly

drprajap: agree with that, but this free is incorrectly trying to free device side memory from IpcContext object (Device Context which is used in Device side APIs) so if this is intended to free host only ipcImpl_.ipc_bases , it needs to happen in IpcHostContext destructor and not here.

avinashkethineedi: it does not free those device pointer, it just frees the [host memory](https://github.com/ROCm/rocSHMEM/pull/200/commits/c305c7bd49d1c663aeb8a967631ebed4997fb0c3#diff-d140135af3c6b2af206a51d7774298f72476b6dc4b01c5f6cf127f718eab5a6bR46) allocated to hold those device pointers

avinashkethineedi (approved):

PR #201, opened 2025-07-17
Cherrypicks

akolliasAMD:
Cherrypicks for 7.0


edgargabriel (approved):

edgargabriel:
actually, apologies, one question: should this cherry-pick include the change to the compile logic such that we compile RO+IPC by default?

PR #203, opened 2025-07-28
[RO] Fix MPI Communicator Check

Yiltan:
We should be checking against `MPI_COMM_NULL` rather than `NULL`.

This doesn't cause an issue for OMPI but there is a problem for MPICH

### DIFF
--- a/src/reverse_offload/queue_proxy.hpp
+++ b/src/reverse_offload/queue_proxy.hpp
@@ -66,7 +66,7 @@ typedef struct queue_element {
   int datatype{-1};
   int PE_root{-1};
   volatile char *status{nullptr};
-  MPI_Comm team_comm{};
+  intptr_t team_comm{};
   union {
     size_t size;
     unsigned long long atomic_value;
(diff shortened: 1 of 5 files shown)


edgargabriel (commented):
I don't think this is sufficient, we actually pass in NULL (not MPI_COMM_NULL) at many instances, see e.g. 

https://github.com/ROCm/rocSHMEM/blob/develop/src/reverse_offload/context_ro_device.cpp#L81
 (there are a few dozen other cases).

abouteiller:
MPICH is using int for MPI_Comm type, so it will complain when compared to NULL. One way to deal with it is to have the comm member be a (intptr_t) as that is compatible with Open MPI and MPICH.

Yiltan:
This was the only other instance I found, but I think it's okay to leave as-is:

https://github.com/ROCm/rocSHMEM/blob/develop/src/rocshmem.cpp#L232

src/reverse_offload/queue_proxy.hpp
```diff
@@ -66,7 +66,7 @@ typedef struct queue_element {
   int datatype{-1};
   int PE_root{-1};
   volatile char *status{nullptr};
-  MPI_Comm team_comm{};
+  intptr_t team_comm{};
```
abouteiller, line 69: Looks good, but I am surprised that you never have to back-cast to MPI_Comm at the use point

Yiltan: You're absolutely correct. I didn't notice an issue with cray-mpich but there was an issue OMPI. The latest commit should resolve this.

I've tested both mpi implementations with the latest commit

abouteiller (approved):

edgargabriel (approved):

PR #204, opened 2025-07-28
Added ability to build for local gpu by env Variable

akolliasAMD:


abouteiller (commented):
The same can be achieved by simply calling `cmake -DBUILD_LOCAL_GPU_TARGET_ONLY=ON`

CMakeLists.txt
```diff
@@ -92,6 +92,45 @@ include(ROCMCheckTargetIds)
 
 rocm_setup_version(VERSION ${VERSION_STRING})
 
+
+## Check for ROCm version
+# Try to establish ROCM_PATH (for find_package)
+#==================================================================================================
+if(NOT DEFINED ROCM_PATH)
```
Yiltan: Does this duplicate this logic, or does it do something distinct?

https://github.com/ROCm/rocSHMEM/blob/develop/cmake/setup_project.cmake#L34

akolliasAMD: Slightly different. This actually sets it if it is not set

Yiltan (dismissed):

CMakeLists.txt
```diff
@@ -92,6 +92,34 @@ include(ROCMCheckTargetIds)
...
+if (rocm_version_matches)
+  set(ROCM_MAJOR_VERSION ${CMAKE_MATCH_1})
+  set(ROCM_MINOR_VERSION ${CMAKE_MATCH_2})
+  set(ROCM_PATCH_VERSION ${CMAKE_MATCH_3})
+
+  message(STATUS "ROCm version: ${ROCM_MAJOR_VERSION}.${ROCM_MINOR_VERSION}.${ROCM_PATCH_VERSION}")
+
+  # Convert the version components to int for comparison
+  math(EXPR ROCM_VERSION "(10000 * ${ROCM_MAJOR_VERSION}) + (100 * ${ROCM_MINOR_VERSION}) + ${ROCM_PATCH_VERSION}")
+  add_definitions("-DROCM_VERSION=${ROCM_VERSION}")
```
abouteiller: Would be better if we have it in rocshmem_config.h so that we can print it from `rocshmem_info`. 

Isn't the ROCM_VERSION available from code in a .h?

cmake/setup_project.cmake
```diff
@@ -31,6 +31,18 @@ set(CMAKE_BUILD_TYPE "Release" CACHE STRING
 ###############################################################################
 # GLOBAL COMPILE FLAGS
 ###############################################################################
+
+# Try to establish ROCM_PATH (for find_package)
+#==================================================================================================
+if(NOT DEFINED ROCM_PATH)
```
abouteiller, line 37: I don't think we need this, as we prepend CMAKE_PREFIX_PATH with ROCM_PATH on line 46, and we use /opt/rocm as a fallback on line 52, and line 88 in the main CMakeList.txt. 

This could actually conflict with using ROCM_ROOT or hip_ROOT (the cmake blessed way) we could report a version different from what has actually been found.

abouteiller (approved):

PR #206, opened 2025-07-31
Import gda_devel back into develop

abouteiller:
This brings gda_devel back into develop

This is a 'minimal changeset' approach, only the parts that are actually required have been back-imported, we may create followup PRs to bring back other desired, but not required changes (e.g., codestyle changes, backend_bc/DISPATCH macro refactor, removal of unused files and functionalities). 

It essentially does the opposite of #164, and as such supersedes it. 

* [x] undo all removal of IPC and RO
* [x] RO and IPC compile
* [x] RO and IPC in a working state
* [x] reconcile backend_bc and DISPATCH in gpu_ib
* [x] gpu_ib compiles
* [x] gpu_ib in a working state on CX7
* [x] gpu_ib in a working state on BNXT
* [x] gpu_ib not implemented features report as such in functional testers
* [x] validate deepep CX7
* [x] validate deepep BNXT
* [x] All PRs made to gda_devel integrated back
* [ ] Track down missing cleanup/free in backend_gda (and possibly also in ipc)
* [x] Reduce memory consumed by wrk_sync_pool, notably for A2A, currently requires  ROCSHMEM_HEAP_SIZE=$((4*1024*1024*1024)) because A2A buffer is 1.3GB with 40 teams #232 
* [x] share the roadmap of not required but desired cleanup and features that would be good to also import as part of followup PRs.

Note:
Short list of what is imported (see merge commit message for the full changeset)

```
Cherry-picked:

[GDA] Query for the correct GID index (#215) 
gda_devel: teams with MPI initalization (#229) 
gda: Disable non-functional tests  (#216) 
rocSHMEM GDA BNXT POC (#213)   
Select MTU based on network config (#214)    
Revert "Only issue a single completion per wavefront (#199)" (#205)       

Squashed commit of the following:

commit 90761d552392ca1f5261fec2e6a08455b0ebc368
Author: Brandon Potter <BKP@users.noreply.github.com>
Date:   Thu Jul 24 14:50:47 2025 -0500

    Only issue a single completion per wavefront (#199)

commit 0056a8a4a7465d520b85c5cb6829ab88783e82f4
Author: Aurelien Bouteiller <aurelien.bouteiller@amd.com>
Date:   Thu Jul 24 14:12:35 2025 -0400

    non-fetching amos are implicit nbi, we do not need the terminal quiet. (#179)

commit 75d1bfe0b0afa5cfd5a7dfae89e9de6f1087e531
Author: Alsop, John <johnathan.alsop@amd.com>
[...]

PR #208, opened 2025-08-01
gda_devel for gfx950

abouteiller:


Yiltan:
Should this also cherry pick #204 , the same bug will be present in gda_devel

abouteiller:
> Should this also cherry pick #204 , the same bug will be present in gda_devel

I pulled in @akolliasAMD changes to the cmake from #182 with a minimal changeset to detect 7.0 see https://github.com/ROCm/rocSHMEM/pull/208/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR111

abouteiller:
The solution implemented here has issues, we will have to pull-in #204 when it is complete and merged

abouteiller:
There is no point in  pursuing this now that #206 is almost ready

PR #211, opened 2025-08-07
rocSHMEM device API changes for bitcode support

drprajap:
ROCSHMEM_CTX_DEFAULT's visibility needs to be explicitly set to expose that as external device global variable in device bitcode.

### DIFF
--- a/include/rocshmem/rocshmem_common.hpp
+++ b/include/rocshmem/rocshmem_common.hpp
@@ -115,7 +115,7 @@ typedef struct {
 /**
  * Shmem default context.
  */
-extern __constant__ rocshmem_ctx_t ROCSHMEM_CTX_DEFAULT;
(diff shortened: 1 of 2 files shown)


Yiltan:
LGTM

- Validated blocking and non-blocking default context tests (60, 61) which are not a part of the CI for both IPC and RO+IPC
- No observable performance difference by using `__device__` rather than `__constant__` for the default ctx tests

Yiltan (approved):

abouteiller (approved):
Waiting on ci pipeline to run

drprajap:
> LGTM
> 
> * Validated blocking and non-blocking default context tests (60, 61) which are not a part of the CI for both IPC and RO+IPC
> * No observable performance difference by using `__device__` rather than `__constant__` for the default ctx tests

Thank you for checking that out.

Yiltan:
CI looks good - the error on gfx942 is a known expected error

drprajap:
> CI looks good - the error on gfx942 is a known expected error

I will proceed with the merge then.

PR #213, opened 2025-08-08
rocSHMEM GDA BNXT POC

Yiltan:
Please see Confluence page for details


Yiltan:
Validated rocSHMEM functional tests on CX7/RoCE - no errors or performance issues.

One minor bug, shown in [8a56a8e](https://github.com/ROCm/rocSHMEM/pull/213/commits/8a56a8e6677f48d7374eb69b69a7dea95e34f797)

edgargabriel:
> Validated rocSHMEM functional tests on CX7/RoCE - no errors or performance issues.
> 
> One minor bug, shown in [8a56a8e](https://github.com/ROCm/rocSHMEM/pull/213/commits/8a56a8e6677f48d7374eb69b69a7dea95e34f797)

actually, that was added in on purpose, otherwise you cannot test the no-MPI path on gda_devel (since the logic to detect PMIX was not ported over). In addition, this should only impact the example_driver, not rocSHMEM itself.

Yiltan:
> > Validated rocSHMEM functional tests on CX7/RoCE - no errors or performance issues.
> > One minor bug, shown in [8a56a8e](https://github.com/ROCm/rocSHMEM/pull/213/commits/8a56a8e6677f48d7374eb69b69a7dea95e34f797)
> 
> actually, that was added in on purpose, otherwise you cannot test the no-MPI path on gda_devel (since the logic to detect PMIX was not ported over). In addition, this should only impact the example_driver, not rocSHMEM itself.

Not the ifdef, I understand that was added on purpose

The define was added in my original commit but it shouldn't be there. It was subsequently removed in the second commit. So the diff compared to `gda_devel` is null

Yiltan:
- Rebased with `gda_devel`
- Updated bnxt code to handle different MTUs [7d425ea](https://github.com/ROCm/rocSHMEM/pull/213/commits/7d425ea51d1cddd541d295e6d95f50b0a3d85aef)
- Fixed MSN table bug [04cc5a6](https://github.com/ROCm/rocSHMEM/pull/213/commits/04cc5a64b001dce100e1e2208ee4ba6f8998b8ef)
- DeepEP runs to completion

src/util.hpp
```diff
@@ -206,6 +208,100 @@ __device__ void gpu_dprintf(const char* fmt, const Args&... args) {
   }
 }
 
+__device__ __forceinline__ void memcpy(void* dst, void* src, size_t size) {
```
abouteiller, line 211: aren't these normally in utils in develop? Not a blocker just asking to plan for merge-back

abouteiller (approved):

abouteiller:
functionally equivalent to cx7 exept missing fadd

src/util.hpp
Yiltan: Yup: https://github.com/ROCm/rocSHMEM/blob/develop/src/util.hpp#L171

They were removed in gda_devel but I needed it for the store of WQEs

PR #214, opened 2025-08-12
Select MTU based on network config

Yiltan:
## Motivation

Not all NICs have the same MTU configuration. Therefore, on some machines having a hardcoded MTU of 4096 causes issues.

## Technical Details
When we query the port info, we obtain the active MTU. We are now setting that as the MTU that the QP will be configured to use.

### DIFF
--- a/src/gpu_ib/bnxt/gda_device_bnxt.cpp
+++ b/src/gpu_ib/bnxt/gda_device_bnxt.cpp
@@ -86,7 +86,7 @@ void GDADevice::change_status_rtr(ibv_qp *qp, dest_info_t *dest, uint8_t port) {
                               | IBV_QP_MIN_RNR_TIMER;
 
   attr.qp_state               = IBV_QPS_RTR;
-  attr.path_mtu               = IBV_MTU_4096;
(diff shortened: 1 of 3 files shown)


src/gpu_ib/bnxt/gda_device_bnxt.cpp
```diff
@@ -96,6 +96,7 @@ void GDADevice::change_status_rtr(ibv_qp *qp, dest_info_t *dest, uint8_t port) {
   attr.ah_attr.sl             = 1;
   attr.ah_attr.is_global      = 1;
   attr.ah_attr.port_num       = port;
+  rtr.exp_qp_attr.path_mtu    = ib_state->portinfo.active_mtu;
```
gaoikawa: rtr -> attr, yes?

Yiltan: Yup thats a typo

src/gpu_ib/gda_device.cpp
gaoikawa, line 1: Remove formatting changes please - thanks.

abouteiller (approved):

PR #217, opened 2025-08-15
Device bitcode support for IPC - Use IPCContext explicitly to link correct device API implementation

drprajap:
## Motivation

During JIT compilation (in case of triton-dist) Device bitcode module is dynamically loaded after host initialization for specific backend is done (IPC for now) . Device bitcode is also built for specific backend and at runtime bitcode cannot infer inherited objects like Context (which is initialized at host init/setup time) so bitcode uses compile time macro like "ENABLE_IPC_BITCODE" to explicitly use specific backend context like IPCContext so that correct IPC backend implementation is linked from bitcode device API. 

## Technical Details

This change is effective for Device bitcode builds and each backend needs to build bitcode with macro like ENABLE_IPC_BITCODE. ENABLE_RO_BITCODE, ENABLE_IB_BITCODE to resolve and link device APIs correctly.

There will be follow-up changes to add scripts to build this Device bitcode library.

## Test Plan

This change has been tested as part of this PR - https://github.com/ByteDance-Seed/Triton-distributed/pull/86/

## Submission Checklist

### DIFF
--- a/src/rocshmem_gpu.cpp
+++ b/src/rocshmem_gpu.cpp
@@ -314,7 +314,7 @@ __device__ int rocshmem_wg_ctx_create(long options, rocshmem_ctx_t *ctx) {
     ctx->team_opaque = reinterpret_cast<TeamInfo *>(ROCSHMEM_CTX_DEFAULT.team_opaque);
     result = device_backend_proxy->create_ctx(options, ctx);
     if(result) {
-      reinterpret_cast<ContextTy *>(ctx->ctx_opaque)->setFence(options);
+      reinterpret_cast<Context *>(ctx->ctx_opaque)->setFence(options);
     }
   }
   __syncthreads();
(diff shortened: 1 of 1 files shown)


drprajap:
any comments on the review after offline discussion?

abouteiller:
We are planning on shipping a single library with all backends enabled at the same time and that will potentially cause a significant refactor and conflict with this approach. But lets fix future problems when we are there.

drprajap:
> We are planning on shipping a single library with all backends enabled at the same time and that will potentially cause a significant refactor and conflict with this approach. But lets fix future problems when we are there.

Agree, I am testing out IB backend so I will have follow-up change to align this for other backends as well. but don't want to block this change because of that as we are waiting on triton-dist integration at least for IPC backend which depends on this change.

avinashkethineedi (approved):

PR #227, opened 2025-08-19
mark rocshmem_set_attr_uniqueid_args param 'uid' as 'const rocshmem_uniqueid_t *

houqi:
## Motivation

mark input only parameter as const, which is more reasonable. also better for tool that parses the header files and auto-generate some pybinding code.

## Technical Details

too simple to explain more

## Submission Checklist


Yiltan:
I'm hesitant with this change as the current API matches other vendor SHMEM implementations. 

Does this same issue with binding generation occur other vendor SHMEM implementations?

Yiltan:
Closing due to inactivity

PR #141, opened 2025-05-26
[SWDEV-534546] Disable building tests in default build

Yiltan:
# Problem

See SWDEV-534546

# Solution

- Disable building tests in default build.
- Enable building of tests in build_config scripts (keep things backwards compatible)

PR #231, opened 2025-08-28
remove `rocthrust` and `rocprim` dependencies from `free_list` unit test

avinashkethineedi:
## Motivation

The `free_list` unit tests previously depended on `rocthrust` and `rocprim`.
These dependencies are unnecessary for the scope of the tests and increase build complexity.

## Technical Details

- Refactored `free_list` unit tests to eliminate use of `rocthrust` and `rocprim`.
- Replaced functionality with standard C++ vectors and device buffers.
- Ensured no loss of coverage or functionality in the tests.

## Test Plan

- Built the unit tests and executed the `free_list` unit tests locally.
- Verified all tests compile and run successfully without `rocthrust` and `rocprim`.

## Test Result

- All unit tests passed successfully.

## Submission Checklist

- [x]  Code compiles locally
- [x] Unit tests pass locally
- [x] CI pipeline passes

### DIFF
--- a/tests/unit_tests/free_list_gtest.cpp
+++ b/tests/unit_tests/free_list_gtest.cpp
@@ -24,8 +24,6 @@
 
 #include "free_list_gtest.hpp"
 
-#include <thrust/sort.h>
-
 #include "../src/util.hpp"
 
 using namespace rocshmem;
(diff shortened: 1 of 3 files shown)


tests/unit_tests/free_list_gtest.cpp
```diff
@@ -76,71 +74,95 @@ TYPED_TEST(FreeListTestFixture, pop_empty_device) {
...
+
+  T *results {nullptr};
+  bool *is_empty {nullptr};
+  size_t size_bytes = sizeof(T) * h_input.size() + sizeof(bool);
+  hip_allocator_.allocate(reinterpret_cast<void**>(&results),
+                          size_bytes);
 
-  thrust::device_vector<T> results(h_input.size());
+  CHECK_HIP(hipMemset(results, 0, size_bytes));
+  is_empty = reinterpret_cast<bool*>(results + h_input.size());
```
Yiltan, line 110: In line 105 we have:

`  size_t size_bytes = sizeof(T) * h_input.size() + sizeof(bool);`

I assume `is_empty` is the last segment after the input in the `results` buffer. 

So in line 110, should the offset be `sizeof(T) * h_input.size()`? I think `std::vector::size` returns the number of elements not the size of the buffer of the vector?

As is, this should should work for some datatypes like `char`

avinashkethineedi: `results` is a type `T*`, so pointer arithmetic happens in steps of `sizeof(T)`. Moving it by `h_input.size()` shifts the pointer by `sizeof(T) * h_input.size()` bytes.

Yiltan (approved):
LGTM

Side note: do we want to revert #141, that was a temporary work around but this PR is the actual solution

PR #232, opened 2025-09-03
Cleanup/remove a2a buff

abouteiller:
## Motivation

We allocate a large amount of memory for Alltoall pWrk buffer that actually is never used for anything. 

## Technical Details

In both IPC and GDA we allocate a large buffer in the pWrk-pSync_pool buffers. In GDA we use heap memory to allocate this buffer and that causes applications to run out of memory because the available heap size is much reduced w.r.t. ROCHSMEM_HEAP_SIZE. 

## Test Plan

Compiled GDA, IPC, RO and ran
`scripts/functional_tests/driver.sh tests/functional_tests/rocshmem_functional_tests all logs ~/hostfile2-ppn8`

`scripts/functional_tests/driver.sh tests/functional_tests/rocshmem_functional_tests gda logs-gda ~/hostfile2-ppn8`

## Test Result

Functional tests pass

## Submission Checklist

### DIFF
--- a/include/rocshmem/rocshmem_common.hpp
+++ b/include/rocshmem/rocshmem_common.hpp
@@ -88,7 +88,6 @@ typedef struct {
 } rocshmem_team_config_t;
 
 constexpr size_t ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE = 1024;
-constexpr size_t ROCSHMEM_ATA_MAX_WRKDATA_SIZE = (4 * 1024 * 1024);
 constexpr size_t ROCSHMEM_BARRIER_SYNC_SIZE = 256;
 constexpr size_t ROCSHMEM_REDUCE_SYNC_SIZE = 256;
 // Internally calls sync function, which matches barrier implementation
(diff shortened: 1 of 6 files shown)


avinashkethineedi:
LGTM, but do we also want to remove `ROCSHMEM_ATA_MAX_WRKDATA_SIZE` from the [header file](https://github.com/ROCm/rocSHMEM/blob/develop/include/rocshmem/rocshmem_common.hpp#L91), since it is not used in any of the backends

avinashkethineedi (dismissed):

abouteiller:
> LGTM, but do we also want to remove `ROCSHMEM_ATA_MAX_WRKDATA_SIZE` from the [header file](https://github.com/ROCm/rocSHMEM/blob/develop/include/rocshmem/rocshmem_common.hpp#L91), since it is not used in any of the backends

Good catch, was this part of our public API or it just got exposed here for convenience?

avinashkethineedi:
> > LGTM, but do we also want to remove `ROCSHMEM_ATA_MAX_WRKDATA_SIZE` from the [header file](https://github.com/ROCm/rocSHMEM/blob/develop/include/rocshmem/rocshmem_common.hpp#L91), since it is not used in any of the backends
> 
> Good catch, was this part of our public API or it just got exposed here for convenience?

OpenSHMEM use to provided compile time constants to specify the length of the work array required for collective routines ([section 6](http://www.openshmem.org/site/sites/default/site_files/OpenSHMEM-1.6.pdf)), but these constants are now deprecated.

Yiltan (approved):

PR #233, opened 2025-09-03
Unify common BNXT and MLX5 initialization code

Yiltan:
## Motivation

We need to eventually select the correct NIC vendor at runtime.

Refactoring the initialization code so that common NIC code can be placed together. This will make future runtime selection easier.

## Technical Details

- Unify `ib_init()` between NIC vendors
  - 1659987d7a5f24e025c77fd52fb106d8998e6c6c
- Refactor and unify QP creation logic
  - Unify QP transition: 678522248e428408539cb642191091d041daab5f
  - QP refactor: 586af6c7706e4f78be5db13579bc892f418e64f8
- Unify CQ logic
  - 816042c9f2eb036096f6b70e2954b2acc8f8d112
- General refactor
  - 850014a4d63dc433e637460e321a2058562ac5df
  - d8e74f2f6d68a01c343a0df79467d9ff8f1b8e48
  - 7befbd8447beb088c22f6dfa4b30c523d0f80148
  - 8f482576284b653ae38399a12ca5d615c8582bfc
  - 9e28530d909ee3dfaf9127a2d35f76d249b4d8ae
  - 3b6fc65d88001789e5a0690f0f6da9684b9b68bb
- Free IB resources on exit
  -  54ff46c7b07c3557a4c842c794f8c30ec7af13e1

## Test Plan

- [x] Validate CX7 IB
- [x] Validate BNXT
- [x] Validate DeepEP


src/gda/queue_pair.cpp
```diff
@@ -37,37 +37,53 @@
...
-  allocator.allocate((void**)&fetching_atomic, 8 * FETCHING_ATOMIC_CNT);
-  CHECK_HIP(hipMemset(fetching_atomic, 0, 8 * FETCHING_ATOMIC_CNT));
-  access = IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE | IBV_ACCESS_REMOTE_READ | IBV_ACCESS_REMOTE_ATOMIC;
-  mr = ibv_reg_mr(pd, fetching_atomic, 8 * FETCHING_ATOMIC_CNT, access);
-  CHECK_NNULL(mr, "ibv_reg_mr");
 #if defined(GDA_IONIC) || defined(GDA_BNXT)
-  fetching_atomic_lkey = mr->lkey;
+  nonfetching_atomic_lkey = mr_nonfetching_atomic->lkey;
+  fetching_atomic_lkey = mr_fetching_atomic->lkey;
 #else
```
edgargabriel: instead of an `#else` could we make this an `# if defined (GDA_IB) ` (or whatever define we have for CX7)

Yiltan: We don't have a directive for CX7 as we assume it to be the default. I'll add a new directive to make the ifdefs simpler. There's a few other places in the code that do this, so I'll add it there too.

abouteiller: GDA_MLX5

src/gda/debug_gda.hpp
```diff
@@ -0,0 +1,237 @@
...
+  DPRINTF("  (void*)    buf             = %p\n",     cq_dv->buf);
+  DPRINTF("  (__be32*)  dbrec           = %p\n",     cq_dv->dbrec);
+  DPRINTF("  (uint32_t) cqe_cnt         = %u\n",     cq_dv->cqe_cnt);
+  DPRINTF("  (uint32_t) cqe_size        = %u\n",     cq_dv->cqe_size);
+  DPRINTF("  (void*)    cq_uar          = %p\n",     cq_dv->cq_uar);
+  DPRINTF("  (uint32_t) cqn             = 0x%x\n",   cq_dv->cqn);
+  DPRINTF("  (uint64_t) comp_mask       = 0x%lx\n",  cq_dv->comp_mask);
+  DPRINTF("================== CQ_DUMP_END ================\n");
+}
+#endif // !GDA_IONIC
```
abouteiller:
```suggestion
#endif // GDA_MLX5
```

src/gda/debug_gda.hpp
```diff
@@ -0,0 +1,237 @@
...
+
+#ifndef LIBRARY_SRC_GDA_DEBUG_GDA_HPP_
+#define LIBRARY_SRC_GDA_DEBUG_GDA_HPP_
+
+static void dump_ibv_context(struct ibv_context *x);
+static void dump_ibv_device(struct ibv_device *x);
+static void dump_ibv_pd(struct ibv_pd *x);
+static void dump_ibv_port_attr(struct ibv_port_attr *x);
+static void dump_ibv_qp(struct ibv_qp *qp, int conn_num);
+static void dump_mlx5dv_qp(struct mlx5dv_qp *qp_dv, int conn_num);
```
abouteiller, line 35: #if defined(GDA_MLX5)

src/gda/queue_pair.cpp
```diff
@@ -37,37 +37,53 @@
...
+  CHECK_NNULL(mr_fetching_atomic, "ibv_reg_mr");
 
-  allocator.allocate((void**)&fetching_atomic, 8 * FETCHING_ATOMIC_CNT);
-  CHECK_HIP(hipMemset(fetching_atomic, 0, 8 * FETCHING_ATOMIC_CNT));
-  access = IBV_ACCESS_LOCAL_WRITE | IBV_ACCESS_REMOTE_WRITE | IBV_ACCESS_REMOTE_READ | IBV_ACCESS_REMOTE_ATOMIC;
-  mr = ibv_reg_mr(pd, fetching_atomic, 8 * FETCHING_ATOMIC_CNT, access);
-  CHECK_NNULL(mr, "ibv_reg_mr");
-#if defined(GDA_IONIC) || defined(GDA_BNXT)
-  fetching_atomic_lkey = mr->lkey;
+#if defined(GDA_MLX5)
```
abouteiller, line 59: What do you think of a wrapper for `htobe32` that performs the swap on device types where this is needed and does nothing otherwise?

src/gda/queue_pair.cpp
```diff
@@ -510,7 +526,7 @@ __device__ uint64_t QueuePair::post_wqe_amo(int pe, int32_t size, uintptr_t *rad
   }
   return ret;
 }
-#else // !GDA_IONIC || !GDA_BNXT
+#else // GDA_MLX5
```
abouteiller:
```suggestion
#if defined(GDA_MLX5)
```

src/gda/backend_gda.cpp
```diff
@@ -1000,7 +1112,7 @@ void GDABackend::init_gid_index(uint8_t port_num) {
 
     current_gid = gid_entries[i].gid;
 
-    err = ibv_query_gid(ctx, port_num, i, &current_gid);
+    err = ibv_query_gid(context, port, i, &current_gid);
```
abouteiller, line 1115: Either way LGTM as long as consistent accross the board

src/gda/queue_pair.cpp
Yiltan: I think that would be a good thing to do. We'd probably have to unify the use of `htobeXX` operations with the `swap_endian` code as they do the same/similar thing. Then from there we could choose to reorder bytes based on the NIC.

I would assume it would be separate PR but can add here

edgargabriel: @Yiltan correct me if I am wrong, but I thought one of the issues is that the flags/fields have to be in little endian for BCOM and AINIC, but in big ending for MLX5? So its not just what the platform supports, but also what the NIC driver requires

Yiltan: I think you have the correct understanding.

I understood Aurelien's comment as adding something line this:

```
uint32_t pack_lkey(uint32_t lkey) {
#if defined(GDA_MLX5)
  return htobe32(lkey);
#else // GDA_IONIC || GDA_BNXT
  return lkey;
#endif
}
```
Then simplify the common code to just call the `pack_lkey` function regardless of NIC type

Yiltan:
Applied comments, rebased with develop, and retested with the above test plan.

abouteiller:
errors in ci pipeline are related to RO, so not caused by this pr

abouteiller (approved):

PR #234, opened 2025-09-03
Remove broken link and minor edits

yugang-amd:
## Submission Checklist

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,7 +6,7 @@
 rocSHMEM documentation
 ****************************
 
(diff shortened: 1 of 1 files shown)


abouteiller:
openshmem.org being unreachable should be temporary. We have reached out to them to verify.

edgargabriel:
openshmem.org is not unreachable. It is an http address instead of https, you need to explicitly add that to the browser link, otherwise it will return a failure.

Yiltan (approved):

PR #235, opened 2025-09-03
Unify environment variable management

omor1:
## Motivation
Environment variables are managed independently in an ad-hoc manner by each subsystem. This results in duplicated code and increased fragility, as well as making it difficult to print out the full configuration of the system.

## Technical Details
- [x] Create new env var config infrastructure with class `rocshmem::config::var<Type, Category>`
- [x] Unify all current env vars under new config
- [x] Add documentation for all env vars
  - Will be done as a separate PR - Update (post merge - was done in #341 and #344)
- [x] Add new `ROCSHMEM_DEBUG` env var

## Test Plan
- [x] Unit tests for `rocshmem::config::var<Type, Category>` and config parsers under `rocshmem::config::parser`

## Test Result
TBD

## Submission Checklist

### DIFF
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -32,6 +32,10 @@
 
 namespace rocshmem {
 namespace config {
+  inline namespace _base {
+    const var<bool> uniqueid_with_mpi("UNIQUEID_WITH_MPI", "", false);
+  }  // inline namespace _base
+
   namespace _detail {
     std::tuple<var_map_t&, std::mutex&> get_var_map() {
       // construct on first use idiom
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -38,6 +38,7 @@ namespace rocshmem {
 namespace config {
   inline namespace _base {
     const var<bool> uniqueid_with_mpi("UNIQUEID_WITH_MPI", "", false);
+    const var<size_t> heap_size("HEAP_SIZE", "", 1L << 30);
     const var<size_t> max_num_host_contexts("MAX_NUM_HOST_CONTEXTS", "", 1);
     const var<size_t> max_num_contexts("MAX_NUM_CONTEXTS", "", 32);
     const var<size_t> max_wavefront_buffers("MAX_WF_BUFFERS", "", 1024);
--- a/src/config.cpp
+++ b/src/config.cpp
@@ -38,6 +38,7 @@ namespace rocshmem {
 namespace config {
   inline namespace _base {
     const var<bool> uniqueid_with_mpi("UNIQUEID_WITH_MPI", "", false);
+    const var<types::debug_level> debug_level("DEBUG_LEVEL", "", types::debug_level::NONE);
     const var<size_t> heap_size("HEAP_SIZE", "", 1L << 30);
     const var<size_t> max_num_teams("MAX_NUM_TEAMS", "", 40);
     const var<size_t> max_num_host_contexts("MAX_NUM_HOST_CONTEXTS", "", 1);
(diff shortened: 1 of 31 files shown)


src/config.hpp
```diff
@@ -0,0 +1,302 @@
...
+    // in particular, the map is allocated so as to fix the static initialization order problem
+    // since these are used inside the constructor for config::var<T, C> to register variables
+    // which are expected to be allocated statically as well
+    std::tuple<var_map_t&, std::mutex&> get_var_map();
+
+    // register the var<T, C> with the global variable map
+    // map from category C to a list of variables in that category
+    // returns a const_iterator to the inserted variable
+    // list is heterogeneous over all valid variable types, using variant<detail::var<T>&...>
+    // locks mutex to ensure that there aren't race conditions due to parallel modifications
```
Yiltan, line 326: Why do we need a mutex?

I assume we would read all environment variable on `rocshmem_init()` which would be single threaded from the host. 

Then at runtime we may read that variable from multiple host threads (i.e RO) or multiple GPU threads. As we are only reading the initial configuration. I don't expect there to be race conditions as `rocshmem_init()` should be called before any other rocSHMEM APIs.

omor1: The intent is that `rocshmem::config::var` instances have static storage duration and are initialized at the start of execution. However, I don't think that there's a way to enforce such a restriction.

My concern was therefore about the situation where someone does something silly like construct env vars at runtime instead, possibly from (multiple) other threads. My thought was that if something like that _does_ happen, it will almost inevitably corrupt the map and/or list(s), causing hard-to-debug issues. My paranoia here is to add the lock around accessing the map; since this is a one-time cost borne only during initialization and exit, there are fairly few env vars, and it should mostly be accessed only by a single thread, that this cost is fairly negligible.

Note that the mutex protects _only_ the map, not the variable itself. The map is used mostly to easily read and print out all the env vars; this particular aspect is not yet implemented. The env var values cannot be modified outside of the class, so after they are constructed there is no issue reading them from multiple threads.

src/config.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <tuple>
+#include <unordered_map>
+
+#include <unistd.h>
+
+namespace rocshmem {
+namespace config {
+  inline namespace _base {
+    const var<bool> uniqueid_with_mpi("UNIQUEID_WITH_MPI", "", false);
+    const var<size_t> max_num_host_contexts("MAX_NUM_HOST_CONTEXTS", "", 1);
```
Yiltan, line 44: This could probably be `int` rather than `size_t`, I don't expect to have more than `INT_MAX` contexts or wavefront buffers. 

Later on in the code we convert pass `size_t` values in to functions with `int` arguments with no casting. I don't expect there to be issues unless we expect the environment variable to be set to a really large number. However, I haven't given that part a thorough review yet

omor1: Both `ROCSHMEM_MAX_NUM_CONTEXTS` and `ROCSHMEM_MAX_WF_BUFFERS` already were using `size_t`; I just kept those the same. `ROCSHMEM_MAX_NUM_HOST_CONTEXTS` used `int`, but I changed it to `size_t` to align with the other two.

Regardless, we'd want the variables to be unsigned, otherwise users can pass in nonsense values like -1 and cause the code to explode without any errors, or we'd need to add our own error checking that would effectively make it unsigned. Since these are logically 'size' values, I'd rather use `size_t` to indicate as such unless there's a real practical or performance issue preventing the type's use, even though we expect the maximum value to be fairly small for all practical uses.

I tried enabling warnings (`-Wall`) to check whether I was introducing anything new, but the existing glut of warnings drowned out anything new. Maybe a future project should be to eliminate as many warnings as possible...

omor1:
Bikeshedding:
* `ROCSHMEM_DEBUG`, like RCCL
* `ROCSHMEM_VERBOSE`
* `ROCSHMEM_LOG_LEVEL`, like UCX

edgargabriel (dismissed):
LGTM. There might be some conflict with the GDA+IPC pr since that renames one of the environment variables, but that can probably easily be resolved.

abouteiller:
I would prefer we merge gda+ipc first if possible :)

LGTM too,  this code may benefit from having unit-testing and functional testing added.

omor1:
Renamed config -> envvar per discussion with @abouteiller. Unit tests will be added soon.

omor1:
c2fe529 prints the error function name using `std::source_location` from C++20. If we want to avoid that, we can use the GCC- (and LLVM)-specific `__PRETTY_FUNCTION__` or the standard `__func__` - but the latter includes only the bare function name. We could also just manually put in the function name as a string in the code, but then it won't include the templated type.

src/team_tracker.hpp
```diff
@@ -93,7 +94,7 @@ class TeamTracker {
    *
    * @return number of teams supported by tracker
    */
-  __host__ __device__ int get_max_num_teams() { return max_num_teams_; }
+  __host__ __device__ int get_max_num_teams() { return max_num_teams_; } // TODO fix narrowing
```
abouteiller, line 97: What's the problem here?

abouteiller (approved):

edgargabriel (approved):
LGTM

src/team_tracker.hpp
omor1: I changed the definition of `max_num_teams` to `size_t`, matching `envvar::max_num_teams`. It is logically a "size", so using `size_t` makes sense even if we never expect the number of teams to be especially large.

However, unlike (most of) the other variables, `max_num_teams` is also used on the device: `get_max_num_teams()` is marked `__host__ __device__`, and I'm not sure what the consequence on correctness and/or performance is of changing the function signature to return `size_t` rather than `int`. At least a couple uses (in `backend_ipc.cpp` and `backend_gda.cpp`) seem to *possibly* rely on the variable being an `int`.

avinashkethineedi: That function is not currently used on the device. It was marked with the `__device__` specifier to allow device-side instantiation, but since that is not supported at the moment, I don’t see any issue with changing it to `size_t`.

avinashkethineedi (approved):

Yiltan (approved):

PR #237, opened 2025-09-05
Remove unused scripts

abouteiller:
## Motivation

Remove older scripts that are not in-use anymore

## Question

Is the driver.py better than driver.sh we use most?


edgargabriel:
I am ok with removing the first two scripts, but I thought we had the agreement that we keep the scripts that generate the header files in the repo, in case we need to regenerate header files for some API changes.

Yiltan:
@edgargabriel, I think that was the initial plan but it seems we haven't really maintained them when we add new APIs. 

For example, the new barrier/sync APIs we added are not present in the python scripts. If we run the scripts it would remove these APIs.

If we opt to keep them we should maintain them when adding new APIs

abouteiller:
We decided to keep the include generators for now, so their removal has been undone from that pr

Yiltan (approved):

PR #239, opened 2025-09-08
Updated docs for ROCm 7.x.x

Yiltan:
Update docs for next release

PR #240, opened 2025-09-09
api: wg_create_ctx w/o options param

abouteiller:
## Motivation

In the gda_devel branch we had a different API for wg_create_ctx w/o the option parameter. DeepEP customer code has been adapted to use that variant. It is best to enable compatibility with both forms going forward.

## Technical Details

Since we cannot just have an implicit initializer due to the order of parameters (only valid Right->Left), we use function overloading. 

## Test Plan

Modified team_infra tester and sync tester to use the optionless variant for test coverage

## Test Result

All tests pass in IPC (Hayabusa)

## Submission Checklist

### DIFF
--- a/docs/api/ctx.rst
+++ b/docs/api/ctx.rst
@@ -11,7 +11,7 @@ Context management routines
 ROCSHMEM_CTX_CREATE
 -------------------
 
-.. cpp:function:: __device__ int rocshmem_wg_ctx_create(int64_t options, rocshmem_ctx_t *ctx)
+.. cpp:function:: __device__ int rocshmem_wg_ctx_create(long options, rocshmem_ctx_t *ctx)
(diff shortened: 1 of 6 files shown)


Yiltan (approved):
LGTM

Do we want to update the CHANGELOG.md as a part of this PR or should I add it #239

abouteiller:
Easier to have it in #239 to avoid merge conflict

edgargabriel (approved):

abouteiller:
This approach unfortunately does not remove the requirement of changing deepEP code to adapt to APIs (due to how the shmem_wrapper is implemented in their code).

PR under evaluation for value outside initial use case.

abouteiller:
Lets close for now, we can reopen if we revisit that idea

PR #247, opened 2025-09-10
[GDA] implement rocshmem_p

Yiltan:
## Motivation

We were missing `rocshmem_p` functionality for GDA

AIROCSHMEM-161

## Technical Details

- Implemented inline write operations for mlx5 and bnxt
- Enabled "p" tests
- Enable "pingpong" tests which were dependent on `rocshmem_p`

## Test Plan

- [x] Validate CX7
- [x] Validate Thor 2

### DIFF
--- a/src/gda/bnxt/queue_pair_bnxt.cpp
+++ b/src/gda/bnxt/queue_pair_bnxt.cpp
@@ -247,17 +247,26 @@ __device__ void QueuePair::post_wqe_rma(int pe, int32_t length, uintptr_t *laddr
       uint32_t wqe_size;
       uint32_t wqe_type;
       uint32_t hdr_flags;
-      uint32_t rma_slots = 3; // (Three slots: hdr, rdma)
+      uint32_t inline_msg;
+
+      uint32_t rma_slots  = 3; // (Three slots: hdr, rdma, sge)
+
+      inline_msg = length <= inline_threshold &&
+                   opcode == GDA_OP_RDMA_WRITE;
 
       hdr_ptr  = (struct bnxt_re_bsqe*) bnxt_re_get_hwqe(&sq, 0);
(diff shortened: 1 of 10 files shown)


src/gda/bnxt/queue_pair_bnxt.cpp
```diff
@@ -247,17 +247,26 @@ __device__ void QueuePair::post_wqe_rma(int pe, int32_t length, uintptr_t *laddr
...
 
       /* Populate Header Segment */
-      wqe_size  = BNXT_RE_HDR_WS_MASK    & rma_slots;
+      wqe_type  = BNXT_RE_HDR_WT_MASK & opcode;
+      wqe_size  = BNXT_RE_HDR_WS_MASK & rma_slots;
       hdr_flags = ((uint32_t) BNXT_RE_HDR_FLAGS_MASK)
                 & ((uint32_t) BNXT_RE_WR_FLAGS_SIGNALED);
-      wqe_type  = BNXT_RE_HDR_WT_MASK    & opcode;
+
+      if (inline_msg) {
```
Yiltan, line 267: Are we okay with these if/else statements or create two functions: `post_wqe_rma` and `post_wqe_rma_inl`? Same goes for mlx5 code

abouteiller: do we see a variation in message rate?

Yiltan: For mlx5 not really, for bnxt yes. I think thats enough motivation to have two separate kernels

abouteiller: Ok, I think we can defer the split to a later PR

abouteiller (approved):

PR #248, opened 2025-09-10
move dv functionality to use dlopen

edgargabriel:
## Motivation

abstract out the usage of direct verbs functionality to use tables with the function pointers. This will allow in a second step the library to be simultaniously be compiled for multiple NIC vendors/DV libraries and interfaces. For now, the conversion has been done for IB MLX5 and BCOM DV, the Pensando AINIC is to follow soon.

## Technical Details
dlopen libbnxt_re.so and look for the functions used for BCOM Thor2 NICs
dlopen libmlx5.so and look for the functions used for MLX5 CX7 NICs

## Test Plan

- [x] Code compiles locally
- [x]  Validate BCOM Thor2 
- [x]  Validate IB CX7

### DIFF
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -71,6 +71,29 @@ GDABackend::GDABackend(TcpBootstrap *bootstrap):  Backend(bootstrap) {
 
 void GDABackend::init() {
   type = BackendType::GDA_BACKEND;
+  int ret;
+
+#if defined(GDA_BNXT)
+  ret = bnxt_dv_dl_init();
+  if (ret != ROCSHMEM_SUCCESS) {
+    // Disable BNXT GDA support.
+    DPRINTF("Initializing rocSHMEM BNXT GDA support failed\n");
+    // We abort for now, but might remove that once we support
+    // multiple NIC types in the same build
+    abort();
+  }
+#endif
+#if defined(GDA_MLX5)
+  ret = mlx5_dv_dl_init();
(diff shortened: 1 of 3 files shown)


src/gda/backend_gda.cpp
```diff
@@ -113,6 +136,15 @@ GDABackend::~GDABackend() {
   cleanup_gpu_qps();
   cleanup_heap_memory_rkey();
   cleanup_ibv();
+
+#if defined(GDA_BNXT)
+  if (bnxtdv_handle_ != nullptr)
```
Yiltan, line 141: Maybe `bnxt/mlx5_dv_dl_finalize()` to make APIs symmetric but its not needed

Yiltan (approved):
Validated on thor2/cx7 no observable issues

abouteiller (approved):

edgargabriel:
one thing I havne't done yet is to remove the mlx5 and bnxt_re libraries from the dependencies. Should I try to add that as well?

src/gda/backend_gda.cpp
edgargabriel: I was thinking about it, because its a single line of what we need to execute, I felt like we could leave it as is, but I am open to change it in a subsequent PR.

edgargabriel:
> one thing I havne't done yet is to remove the mlx5 and bnxt_re libraries from the dependencies. Should I try to add that as well?

Decided to fix this in a later PR.

PR #250, opened 2025-09-11
Make it possible to test RMA GET and PUT separately

abouteiller:
DISABLE_GET removed from ALL, idea is that the CI scripts will invoke a subset that is known to work rather than ALL

## Motivation

We need to have granular testing capabilities. Here we can trigger Put and Get separately. This may prove useful for CI scripts as different transports have a different set of features implemented.

Idea is that CI scripts would run explicitly the features that are known working rather than calling all. 

## Technical Details

In driver.sh: Remove DISABLE_GET, add a 'get' (and 'put') command line option so one can run put and get separately.

## Test Plan

No test plan, this presumably will need adaptations in CI scripts for RO due to removal of DISABLE_GET

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -286,9 +286,7 @@ TestRMAGet() {
 
 TestRMA() {
   TestRMAPut
-  if [ "0" == "$ROCSHMEM_DRIVER_DISABLE_GET" ]; then
-    TestRMAGet
-  fi
+  TestRMAGet
 }
 
 TestAMO() {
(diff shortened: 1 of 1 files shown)


edgargabriel (approved):

Yiltan:
I'm pretty sure this will break gfx90a pipelines. "all" should change TestRMA to TestPut to avoid this.

Sorry, i should have caught this before merge

abouteiller:
No, I was trigger happy. Edgar is modifying the pipelines on the Jenkins side so that should be fixed rapidly.

edgargabriel:
then let me do the following tomorrow: I will split out the jenkins modifications into two separate PRs:

- one is to separate the tests executed for IPC and RO + add the non-MPI testpath for the IPC conduit
- make the changes requiring for the gda testing (which is more intrusive), 

This way we can fix the CI quickly, and extend it when its ready

PR #251, opened 2025-09-11
Fix broken atomics from PR #233

Yiltan:
## Motivation

Atomics stoped working

## Technical Details

* The QueuePair object was out of scope at the end of the for loop. So the deconstructor was called.
* Although correct for C++ to do this, it ignores that we copied the QueuePair object into device memory and have an instance there.
* Early deconstruction resulted in calling ibv_dereg_mr on the atomics memory region. So when the GPU kernel tried to use the memory region it wasn't registered which resulted in a protection domain error.
* The solution was to allocate our QueuePair obj with the new operator which leaves memory management to us, then we can manually call the deconstructor.

## Test Plan

Test using the driver script.

Note: atomic tests remain disabled as it will fail as we are yet to support 32bit atomics:q

## Test Result

- [x] Validated CX7
- [x] Validated Thor2

### DIFF
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -670,16 +670,35 @@ void GDABackend::cleanup_heap_memory_rkey() {
 }
 
 void GDABackend::setup_gpu_qps() {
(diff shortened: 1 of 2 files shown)


abouteiller (approved):

abouteiller:
validated deepEP internode Cx/7 and bnxt

PR #252, opened 2025-09-11
add an all-ro flag

edgargabriel:
to specify the subset of tests that we want to run in Jenkins with the RO conduit

## Motivation

simplifies testing of the relevant functionality for the CI


edgargabriel:
(was wondering whether the noCI flag makes sense, but I think it does, as long as the jenkins script does not use this flag, its pointless to run CI for it, there is no code change associated with it).

abouteiller (approved):

PR #253, opened 2025-09-12
[GDA] Implement fetching atomics for BNXT

Yiltan:
## Motivation

Fetching atomics were not implemented for broadcom

## Technical Details

- Fetch into a buffer and return that value
- Indented get tests in driver script

## Test Plan

- [x] Validate on BNXT


src/gda/bnxt/queue_pair_bnxt.cpp
```diff
@@ -363,8 +364,16 @@ __device__ uint64_t QueuePair::post_wqe_amo(int pe, int32_t length, uintptr_t *r
       amo.swp_dt = atomic_data;
 
       /* Populate SG Segment - (Return address of atomic) */
-      sge.pa     = (uint64_t) nonfetching_atomic;
-      sge.lkey   = nonfetching_atomic_lkey;
+      if (fetching) {
+        atomic_idx = __hip_atomic_fetch_add(&fetching_atomic_idx, 1, __ATOMIC_SEQ_CST, __HIP_MEMORY_SCOPE_AGENT);
+        atomic_idx = atomic_idx % FETCHING_ATOMIC_CNT;
+
+        sge.pa     = (uint64_t) &fetching_atomic[atomic_idx];
```
avinashkethineedi, line 369: multiple threads might end-up using the same memory location simultaneously in the default context

Yiltan: https://github.com/ROCm/rocSHMEM/blob/develop/src/gda/bnxt/provider_gda_bnxt.hpp#L34

We set the max atomics to 1, so we can only service one atomic at a time per QP

abouteiller (changes requested):
You have a .swp file that ended up committed :)

src/gda/bnxt/queue_pair_bnxt.cpp
```diff
@@ -391,6 +398,10 @@ __device__ uint64_t QueuePair::post_wqe_amo(int pe, int32_t length, uintptr_t *r
     release_lock(&sq.lock);
   }
 
+  if (fetching) {
+    return fetching_atomic[atomic_idx];
```
abouteiller, line 402: https://github.com/ROCm/rocSHMEM/pull/253/files#diff-2548ac2b0900f38b49bf1847b485ba1f7d535eef55217960acb02e291c0a1a1dR394

could we move that threadfence and quiet in the fetching if here?

abouteiller (approved):
Could

src/gda/bnxt/queue_pair_bnxt.cpp
Yiltan: I think we could, I'll look into it and create another PR for that

PR #260, opened 2025-09-19
Cleanup/wg init

abouteiller:
## Motivation

In gda_devel we had removed these functions with no adverse effect. These are not part of the OpenSHMEM spec.

## Technical Details

1. Remove calls from tests
2. Remove calls from examples
3. Mark function wg_init, wg_init_thread and wg_finalize as deprecated

## Test Plan

* [x] compile and run functional tests on IPC

### DIFF
--- a/include/rocshmem/rocshmem.hpp
+++ b/include/rocshmem/rocshmem.hpp
@@ -374,7 +374,7 @@ __host__ void rocshmem_global_exit(int status);
  *
  * @return void.
  */
-__device__ void rocshmem_wg_init();
+[[deprecated]] __device__ void rocshmem_wg_init();
 
 /**
  * @brief Finalizes device-side rocSHMEM resources. Must be called before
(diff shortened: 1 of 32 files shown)


include/rocshmem/rocshmem.hpp
```diff
@@ -360,7 +360,7 @@ __host__ void rocshmem_global_exit(int status);
  *
  * @return void.
  */
-__device__ void rocshmem_wg_init();
+[[deprecated]] __device__ void rocshmem_wg_init();
```
Yiltan, line 377: I would also add a deprecation notice in the docs:

https://github.com/ROCm/rocSHMEM/blob/develop/docs/api/init.rst?plain=1#L26

And leave a notice in the changelog, so we don't forget to add the depreciation notice there:

https://github.com/ROCm/rocSHMEM/blob/develop/CHANGELOG.md

Same for other APIs

edgargabriel (dismissed):

Yiltan (dismissed):
LGTM

BTW I updated the docs and changelog to reflect this change

edgargabriel (approved):

Yiltan (approved):

PR #261, opened 2025-09-19
Select host NIC vendor code at runtime

Yiltan:
## Motivation

- Select host NIC vendor code at runtime is required so we can provide a single package

## Technical Details

- Using dlopen we select the nic vendor type
- Using the nic vendor type, I have removed most of the ifdefs in the host code
   - IONIC ifdefs cannot be removed as we do not have dlopen for that nic nor a public header
 
### Notes
- To remove the remaining BXNT ifdefs, we would need to refactor the dlopen code so that the NIC vendor can be select in rocshmem_init and we create the correct backend class.
- This will remain for another PR as there is on going work with dlopen

## Test Result

- [x] Validate functional tests MLX5
- [x] Validate functional tests BNXT

### DIFF
--- a/cmake/FindIBVerbs.cmake
+++ b/cmake/FindIBVerbs.cmake
@@ -48,24 +48,24 @@ find_library(IBVerbs_PROVIDER_LIBRARY
   HINTS ${PC_IBVerbs_LIBDIR} ${PC_IBVerbs_LIBRARY_DIRS}
   PATH_SUFFIXES lib lib64
 )
-elseif (GDA_BNXT)
-find_library(IBVerbs_PROVIDER_LIBRARY
-  NAMES bnxt_re libbnxt_re
-  HINTS ${PC_IBVerbs_LIBDIR} ${PC_IBVerbs_LIBRARY_DIRS}
-  PATH_SUFFIXES lib lib64
-)
-else()
-find_library(IBVerbs_PROVIDER_LIBRARY
-  NAMES mlx5 libmlx5
-  HINTS ${PC_IBVerbs_LIBDIR} ${PC_IBVerbs_LIBRARY_DIRS}
-  PATH_SUFFIXES lib lib64
-)
-endif()
 
 find_package_handle_standard_args(IBVerbs DEFAULT_MSG
   IBVerbs_LIBRARY IBVerbs_INCLUDE_DIR IBVerbs_PROVIDER_LIBRARY
 )
 mark_as_advanced(IBVerbs_LIBRARY IBVerbs_INCLUDE_DIR IBVerbs_PROVIDER_LIBRARY)
+
+add_library(IBVerbs::verbs_provider UNKNOWN IMPORTED)
+set_target_properties(IBVerbs::verbs_provider PROPERTIES
+  IMPORTED_LOCATION "${IBVerbs_PROVIDER_LIBRARY}"
+  INTERFACE_INCLUDE_DIRECTORIES "${IBVerbs_PROVIDER_INCLUDE_DIR}"
+)
+target_link_libraries(IBVerbs::verbs_provider)
+endif()
+
+find_package_handle_standard_args(IBVerbs DEFAULT_MSG
+  IBVerbs_LIBRARY IBVerbs_INCLUDE_DIR
+)
+mark_as_advanced(IBVerbs_LIBRARY IBVerbs_INCLUDE_DIR)
 
 if (IBVerbs_FOUND)
 add_library(IBVerbs::verbs UNKNOWN IMPORTED)
@@ -74,10 +74,7 @@ set_target_properties(IBVerbs::verbs PROPERTIES
   INTERFACE_COMPILE_OPTIONS "${PC_IBVerbs_CFLAGS_OTHER}"
   INTERFACE_INCLUDE_DIRECTORIES "${IBVerbs_INCLUDE_DIR}"
 )
-add_library(IBVerbs::verbs_provider UNKNOWN IMPORTED)
-set_target_properties(IBVerbs::verbs_provider PROPERTIES
-  IMPORTED_LOCATION "${IBVerbs_PROVIDER_LIBRARY}"
-  INTERFACE_INCLUDE_DIRECTORIES "${IBVerbs_PROVIDER_INCLUDE_DIR}"
-)
-target_link_libraries(IBVerbs::verbs INTERFACE IBVerbs::verbs_provider)
+
+target_link_libraries(IBVerbs::verbs INTERFACE)
+
 endif()
--- a/cmake/FindIBVerbs.cmake
+++ b/cmake/FindIBVerbs.cmake
@@ -59,7 +59,7 @@ set_target_properties(IBVerbs::verbs_provider PROPERTIES
   IMPORTED_LOCATION "${IBVerbs_PROVIDER_LIBRARY}"
   INTERFACE_INCLUDE_DIRECTORIES "${IBVerbs_PROVIDER_INCLUDE_DIR}"
 )
-target_link_libraries(IBVerbs::verbs_provider)
+target_link_libraries(IBVerbs::verbs IBVerbs::verbs_provider)
 endif()
 
 find_package_handle_standard_args(IBVerbs DEFAULT_MSG
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -75,30 +75,30 @@ void GDABackend::init() {
   type = BackendType::GDA_BACKEND;
 
 #ifdef GDA_IONIC
-  nic_type = GDA_NIC::IONIC;
+  gda_vendor = GDAVendor::IONIC;
 #endif
 
-  if (nic_type == GDA_NIC::INVALID) {
+  if (gda_vendor == GDAVendor::NONE) {
     ret = bnxt_dv_dl_init();
 
     if (ret == ROCSHMEM_SUCCESS) {
-      nic_type = GDA_NIC::BNXT;
+      gda_vendor = GDAVendor::BNXT;
     } else {
       DPRINTF("Initializing rocSHMEM BNXT GDA support failed\n");
     }
   }
 
-  if (nic_type == GDA_NIC::INVALID) {
+  if (gda_vendor == GDAVendor::NONE) {
     ret = mlx5_dv_dl_init();
 
     if (ret == ROCSHMEM_SUCCESS) {
-      nic_type = GDA_NIC::MLX5;
+      gda_vendor = GDAVendor::MLX5;
     } else {
       DPRINTF("Initializing rocSHMEM MLX5 GDA support failed\n");
     }
   }
 
-  if (nic_type == GDA_NIC::INVALID) {
+  if (gda_vendor == GDAVendor::NONE) {
     printf("Initializing rocSHMEM with IONIC, BNXT, or MLX5 GDA support failed\n");
     abort();
   }
@@ -622,7 +622,7 @@ void GDABackend::setup_ibv() {
 void GDABackend::cleanup_ibv() {
   int err;
 
-  if (nic_type == GDA_NIC::BNXT) {
+  if (gda_vendor == GDAVendor::BNXT) {
     CHECK_HIP(hipHostUnregister(db_region_attr.dbr));
 
     for (int i = 0; i < qps.size(); i++) {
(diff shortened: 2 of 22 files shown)


cmake/FindIBVerbs.cmake
```diff
@@ -48,24 +48,24 @@ find_library(IBVerbs_PROVIDER_LIBRARY
...
-  HINTS ${PC_IBVerbs_LIBDIR} ${PC_IBVerbs_LIBRARY_DIRS}
-  PATH_SUFFIXES lib lib64
+mark_as_advanced(IBVerbs_LIBRARY IBVerbs_INCLUDE_DIR IBVerbs_PROVIDER_LIBRARY)
+
+add_library(IBVerbs::verbs_provider UNKNOWN IMPORTED)
+set_target_properties(IBVerbs::verbs_provider PROPERTIES
+  IMPORTED_LOCATION "${IBVerbs_PROVIDER_LIBRARY}"
+  INTERFACE_INCLUDE_DIRECTORIES "${IBVerbs_PROVIDER_INCLUDE_DIR}"
 )
+target_link_libraries(IBVerbs::verbs_provider)
```
abouteiller:
```suggestion
target_link_libraries(IBVerbs::verbs IBVerbs::verbs_provider)
```

src/gda/bnxt/backend_gda_bnxt.cpp
```diff
@@ -218,32 +218,4 @@ void GDABackend::create_qps(int sq_length) {
     CHECK_NNULL(qps[i], "bnxt_re_dv_create_qp");
   }
 }
-
-int GDABackend::bnxt_dv_dl_init() {
```
abouteiller, line 222: is it really required to move it to the generic gda_backend?

src/gda/queue_pair.cpp
```diff
@@ -56,17 +54,35 @@ QueuePair::QueuePair(struct ibv_pd* pd) {
...
     fetching_atomic_freelist->push_back(fetching_atomic + i);
   }
+
+  /* Set Correct opcodes for each NIC */
+#ifdef GDA_IONIC
+  gda_op_rdma_write = IONIC_V2_OP_RDMA_WRITE;
+  gda_op_atomic_fa  = IONIC_V2_OP_ATOMIC_FA;
+  gda_op_atomic_cs  = IONIC_V2_OP_ATOMIC_CS;
+#endif
+  if (nic_type == GDA_NIC::BNXT) {
```
abouteiller: can these be const?

Yiltan: I don't think so as const has to be assigned at compile time

src/gda/bnxt/backend_gda_bnxt.cpp
Yiltan: Not really, I added it hear to remove some additional logic.

After discussion, people are fine with the if else statements here. I'll return it and remove the remaining bnxt ifdefs

src/gda/backend_gda.hpp
```diff
@@ -96,6 +91,13 @@ class GDAHostContext;
 class QueuePair;
 class HostInterface;
 
+enum GDA_NIC {
```
omor1:
```suggestion
enum class GDA_NIC {
```
can maybe bikeshed naming, but don't really know if e.g. `GDAType`, `GDAVendor`, etc. is any better

src/gda/backend_gda.hpp
```diff
@@ -96,6 +91,13 @@ class GDAHostContext;
...
+  INVALID,
```
omor1: Is `INVALID` the best name for this? It seems to be used as a sort of 'auto-select' mechanism that tries each one progressively. If there's a need for a type tag representing that there is no valid GDA device, I'd recommend `NONE`.
```suggestion
  NONE,
  AUTO,
```

src/gda/backend_gda.cpp
```diff
@@ -70,29 +70,38 @@ GDABackend::GDABackend(TcpBootstrap *bootstrap):  Backend(bootstrap) {
...
+      nic_type = GDA_NIC::MLX5;
+    } else {
+      DPRINTF("Initializing rocSHMEM MLX5 GDA support failed\n");
+    }
+  }
+
+  if (nic_type == GDA_NIC::INVALID) {
+    printf("Initializing rocSHMEM with IONIC, BNXT, or MLX5 GDA support failed\n");
     abort();
   }
```
omor1: Can this be put into a separate function? `GDABackend::autodetect_nic` or something.

Does it make sense to do this detection prior to instantiating `GDABackend`? That way `nic_type` can be `const`, passed to `GDABackend::GDABackend`, or possibly even a constant expression via e.g. a template parameter (though that maybe overcomplicates things).

src/gda/backend_gda.hpp
```diff
@@ -125,21 +128,23 @@ class GDABackend : public Backend {
...
   std::vector<dest_info_t> dest_info;
 
-#ifdef GDA_BNXT
+  /* GDA_BNXT START */
   std::vector<struct bnxt_host_qp> bnxt_qps;
   std::vector<struct bnxt_host_cq> bnxt_cqs;
 
   struct bnxt_re_dv_db_region_attr db_region_attr;
-#else
+  /* GDA_BNXT END */
```
omor1, line 136: I don't especially like that regardless of which NIC is actually used, `GDABackend` always has to provide all members needed for all supported NIC types. (See also below e.g. `mlx5dv_ftable_`, `mlx5dv_handle_`.) It's not too bad right now, but if we continue supporting additional vendors in the future this could become a real pain point, especially with support for all the different NICs mixed in the same code.

Possible options:
- keep as is
  - deal with it later if it becomes a problem
  - ... but then we might need to care more about backwards-compatibility
- subclass `GDABackend` for each NIC type, customize behavior in the subclasses
- something like [pointer to implementation](https://cppreference.com/w/cpp/language/pimpl.html): NIC-specific type holding their info, `GDABackend` only has pointer to it
  - `void *` (or pointer to common base type) and cast to the correct type depending on `nic_type` (is this essentially just `dynamic_cast`?)
  - pointer to base type, dynamically dispatched (`virtual`) customization points
- `std::variant` as a type-safe discriminated union
- plain old `union` (implicitly tagged via `nic_type`)

src/gda/backend_gda.cpp
```diff
@@ -818,10 +848,16 @@ void GDABackend::modify_qps_init_to_rtr() {
...
   attr.qp_state               = IBV_QPS_RTR;
   attr.path_mtu               = portinfo.active_mtu;
-  attr.max_dest_rd_atomic     = GDA_MAX_ATOMIC;
   attr.min_rnr_timer          = 12;
   attr.ah_attr.port_num       = port;
 
+  if (nic_type == GDA_NIC::IONIC) {
+    attr.max_dest_rd_atomic = 15;
+  }
+  else if (nic_type == GDA_NIC::BNXT || nic_type == GDA_NIC::MLX5) {
```
omor1:
```suggestion
  } else {
```
`attr.max_dest_rd_atomic = 1` seems to be the default value for the supported NIC types, only IONIC is different

src/gda/backend_gda.cpp
```diff
@@ -863,11 +899,17 @@ void GDABackend::modify_qps_rtr_to_rts() {
...
   attr.qp_state      = IBV_QPS_RTS;
-  attr.max_rd_atomic = GDA_MAX_ATOMIC;
   attr.timeout       = 14;
   attr.retry_cnt     = 7;
   attr.rnr_retry     = 7;
 
+  if (nic_type == GDA_NIC::IONIC) {
+    attr.max_dest_rd_atomic = 15;
+  }
+  else if (nic_type == GDA_NIC::BNXT || nic_type == GDA_NIC::MLX5) {
```
omor1:
```suggestion
  } else {
```

src/gda/queue_pair.cpp
omor1: `const` doesn't need to be assigned at compile time, but it _does_ need to be initialized in the constructor's member initializer list:
```C++
static constexpr uint8_t QueuePair::rdma_write_op(GDA_NIC nic_type) {
  switch (nic_type) {
  case NIC_TYPE::IONIC:
    return IONIC_V2_OP_RDMA_WRITE;
  case NIC_TYPE::BNXT:
    return BNXT_RE_WR_OPCD_RDMA_WRITE;
  case NIC_TYPE::MLX5:
    return MLX5_OPCODE_RDMA_WRITE;
  default:
    abort();
  }
  return 0;
}

QueuePair::QueuePair(struct ibv_pd* pd, int nic_type)
    : gda_op_rdma_write{rdma_write_op(nic_type)},
      gda_op_rdma_read{rdma_read_op(nic_type)},
      gda_op_atomic_fa{atomic_fa_op(nic_type)},
      gda_op_atomic_cs{atomic_cs_op(nic_type)} {
  // etc.
}
```

omor1 (commented):
I'm not certain how maintainable this approach will be if we need to add support for more NIC types or more complex functionality. It's certainly simpler for right now though.

src/gda/backend_gda.cpp
Yiltan: I kept it explicit so we wouldn't forget but this is pretty valid

src/gda/queue_pair.cpp
Yiltan: While this does allow us to use a const. Is adding and maintaining four functions simpler than a single if statement?

src/gda/backend_gda.hpp
Yiltan: Yup, I should adjust the name to be a little more consistent with the code base

src/gda/backend_gda.hpp
Yiltan: Pretty valid point, but it could be a lot of work given some time constraints. Ping me and lets chat offline

src/gda/backend_gda.cpp
Yiltan: I agree with the const comment but I will defer that for another PR. Edgar is working on some dlopen functionality and I don't want to create merge conflicts rn.

src/gda/queue_pair.cpp
```diff
@@ -56,17 +54,35 @@ QueuePair::QueuePair(struct ibv_pd* pd) {
...
+  gda_op_rdma_write = IONIC_V2_OP_RDMA_WRITE;
+  gda_op_atomic_fa  = IONIC_V2_OP_ATOMIC_FA;
+  gda_op_atomic_cs  = IONIC_V2_OP_ATOMIC_CS;
+#endif
+  if (gda_vendor == GDAVendor::BNXT) {
+    gda_op_rdma_write = BNXT_RE_WR_OPCD_RDMA_WRITE;
+    gda_op_rdma_read  = BNXT_RE_WR_OPCD_RDMA_READ;
+    gda_op_atomic_fa  = BNXT_RE_WR_OPCD_ATOMIC_FA;
+    gda_op_atomic_cs  = BNXT_RE_WR_OPCD_ATOMIC_CS;
+  } else {
```
edgargabriel: Instead of an else part here, could we make it 

```
if (vendor == GDAVendor::MLX5) {
```

abouteiller: would switch statements be faster in general that if-elseif chains?

Yiltan: @edgargabriel, @omor1 requested else rather than the explicit condition, so I think we need a rocSHMEM "style guide" or some consensus on how we all want to do it in the code.

@abouteiller, maybe but the constructor is executed on the host so this isn't on the critical path. Not sure how much impact it would make for the initialization cost.

edgargabriel: I would suggest to distinguish between two different scenarios:
 - if the block part contains constant that are clearly bound to a single vendor (as is the case here in the else part due to the usage of `MLX5_OPCODE_RDMA_WRITE` and similar constants), I would suggest to encapsulate it in an vendor specific if-block, not in a generic else block
 - if the block contains generic verbs code, its a valid fallback mechanism that could apply to multiple vendors. I am ok with an else block in that case

src/gda/queue_pair.cpp
omor1: > While this does allow us to use a const. Is adding and maintaining four functions simpler than a single if statement?

Not sure if it's simpler; I was just showing how it could be done.
The functions are there only because it's simpler than having some large cascading ternary operator chain: `nic_type == NIC_TYPE::IONIC ? IONIC_V2_OP_RDMA_WRITE : nic_type == NIC_TYPE::BNXT ? BNXT_RE_WR_OPCD_RDMA_WRITE : nic_type == NIC_TYPE::MLX5 ? MLX5_OPCODE_RDMA_WRITE : (abort(), 0)` There might be some sort of nifty mechanism to simplify this by using a map or an array or something.

src/gda/queue_pair.cpp
omor1: > would switch statements be faster in general that if-elseif chains?

I think it's mostly a wash either way, and in any case this is executed as part of initialization and isn't in the critical path. I'd err on the side of whatever is more readable & maintainable.

abouteiller (approved):

PR #263, opened 2025-09-23
Select device NIC vendor code at runtime

abouteiller:
## Motivation

For packaging reasons we need to have all networks compiled in the release library.

## Technical Details

This goes on top of PR #261 

Add a wrapper switchyard function around `quiet` `post_wqe_rma` `post_wqe_amo` and give distinct names for internal implementation details like `ring_doorbell`

## Test Plan

* [x] Compile with GDA_BNXT AND GDA_MLX5
* [x] Runs on BNXT (functional tests)
* [x] Runs on MLX5 (functional tests)
* [x] Compare performance with monolithic compiled 1-nic solution

### DIFF
--- a/src/gda/bnxt/queue_pair_bnxt.cpp
+++ b/src/gda/bnxt/queue_pair_bnxt.cpp
@@ -222,7 +222,7 @@ __device__ int QueuePair::poll_cq() {
   return 0;
 }
 
-__device__ void QueuePair::quiet() {
+__device__ void QueuePair::quiet_bnxt() {
   uint64_t active_lane_mask;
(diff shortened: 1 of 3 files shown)


Yiltan (dismissed):
LGTM 

but we may want to modify the functional calls in backend_gda to match this PR

https://github.com/ROCm/rocSHMEM/blob/develop/src/gda/backend_gda.hpp#L372

I.e always use <function_name>_\<vendor> instead of sometimes a prefix and other times a suffix

akolliasAMD (approved):

edgargabriel (approved):

PR #264, opened 2025-09-24
Remove MPI compile-time dependency

edgargabriel:
## Motivation

Replace the MPI functionality in rocSHMEM to be used through dlopen'ed symbols
Based on a customer request to remove compile time dependency on MPI library.

The current logic is that we still look for an external MPI library. If we find an external MPI library, everything is compiled the same way as before (except that rocSHMEM internally dlopen's the MPI library and uses a function table with function pointers for all MPI functionality).

If we do not find an external MPI library, rocSHMEM still compiles correctly. However, neither the functional tests nor the unit tests are compiled in that case, those depend strictly on having an external MPI library.

Most of the examples in the example directory are compiled even when an external MPI library is not found. The only exception is the rocshem_init_attr_test, which requires an external MPI library.

The user can require an external MPI library during compile team by setting the cmake option USE_EXTERNAL_MPI=ON.

JIRA ROCM-10966

## Test Result
- [x] ipc conduit w/ external MPI library
   - compilation
   - functional tests
   - unit tests
- [x] ipc conduit w/o external MPI library 
   - compilation
   - examples
- [x] RO conduit w/ external MPI library
   - compilation
   - functional tests
- [x] gda_mlx5 conduit w/ external MPI library
   - compilation
   - functional tests
- [x] gda_mlx5 conduit w/o external MPI library
   - compilation     
 

## Submission Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ if (NOT BUILD_TESTS_ONLY)
       $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include/rocshmem>   # rocshmem_config.h from rocshmem.hpp
       $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
       $<INSTALL_INTERFACE:include>
-      $<$<BOOL:${USE_EXTERNAL_MPI_HEADERS}>:MPI::MPI_CXX>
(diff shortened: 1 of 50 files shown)


Yiltan (approved):
- [x] gda_bnxt conduit w/ external MPI library
  - compilation
  - functional tests
- [x] gda_bnxt conduit w/o external MPI library
  - compilation

PR #265, opened 2025-09-25
[GDA/BNXT] Obtain SQ head from CQE

Yiltan:
## Motivation

- We should use the CQE to determine where we have posted up to in the SQ
- This allows us to remove atomics for control
- This handles the case when the hardware coalesces multiple CQE

## Technical Details

- `__device__ void poll_cq_until(uint32_t requested_available_slots);`
  - Added to allow us to poll until `requested_available_slots` slots are available in the SQ
  - Not in this PR, but this will allow us to improve the slot acquisition logic by only waiting for the required slots rather than calling quiet and freeing up the entire SQ
  

## Test Plan

- [x] Validated functional test on Thor 2

### DIFF
--- a/src/gda/bnxt/provider_gda_bnxt.hpp
+++ b/src/gda/bnxt/provider_gda_bnxt.hpp
@@ -30,6 +30,8 @@ extern "C" {
 #include "gda/bnxt/bnxt_re_hsi.h"
 }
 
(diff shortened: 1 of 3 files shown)


Yiltan:
Closing this PR, will create a different PR with a larger change set

PR #269, opened 2025-10-02
AIROCSHMEM-169 - gda ionic: restore functionality of ionic gda in rocshmem

ahubbe-amd:
## Motivation

Restore functionality of ionic gda in rocshmem.

## Technical Details

Fix cmake (cherry pick).
Fix QueuPair member name conflict.
Cleanup spin lock, move to util.
Update for compatibility with firmware abi.
Apply first set of performance optimizations.

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.

PR #272, opened 2025-10-07
Runtime selection of IONIC

abouteiller:
## Motivation

IONIC needs to be compiled-in for packaging purposes. This PR enables building IONIC even without the supporting libraries and hardware, and dynamic loading the libraries to identify if the IONIC conduit should be selected.

## Technical Details

Grafted on top of PR #269 (will need rebase when merged)

* [Split ionic code to a subdirectory; dyld libionic](https://github.com/ROCm/rocSHMEM/commit/02a2f34cd671673bfe3e7089318c6448815a6acc)
* [Uniform style for the fntable between ionic and the rest](https://github.com/ROCm/rocSHMEM/commit/ada308c9cd875ea6c57b92fc4534c23f4f2b5e17)
* [Move mlx5 gda conduit to a subdir](https://github.com/ROCm/rocSHMEM/commit/3500960634361e944ffd89aca7343c45e8c9ba51)

## Test Plan

* [x] CX7 builds
* [x] CX7 runs
* [x] BNXT builds
* [x] BNXT runs
* [x] IONIC builds
* [x] IONIC runs
* [x] BNXT+IONIC+CX7 builds
   * [x] runs on BNXT
   * [x] runs on IONIC
   * [x] runs on CX7


src/gda/backend_gda.cpp
```diff
@@ -551,33 +536,37 @@ int GDABackend::backend_can_run() {
...
   }
+#endif //defined(GDA_IONIC)
 
-  /* ToDo: opening ionic DV libraries */
+  /* Try opening mlx5 DV libraries */
+#if defined(GDA_MLX5)
+  handle = mlx5_dv_dlopen();
+  if (handle) {
+    dlclose(handle);
+    return ROCSHMEM_SUCCESS;
```
omor1, line 561: `backend_can_run` just looks for at least one GDA vendor that works so that GDA can be enabled, right? The actual GDA vendor selection logic is elsewhere.

src/gda/backend_gda.cpp
```diff
@@ -1099,123 +1102,18 @@ void GDABackend::create_cqs(int cqe) {
...
+    ionic_initialize_gpu_qp(gpu_qp, conn_num);
+    break;
+  case GDAVendor::BNXT:
+    bnxt_initialize_gpu_qp(gpu_qp, conn_num);
+    break;
+  case GDAVendor::MLX5:
+    mlx5_initialize_gpu_qp(gpu_qp, conn_num);
+    break;
+  default:
+    assert(false /* GDAVendor initialize_gpu_qp */);
```
omor1, line 1116:
```suggestion
  switch (gda_vendor) {
  case GDAVendor::NONE:
    assert(false /* GDAVendor initialize_gpu_qp */);
#ifdef NDEBUG
    __builtin_unreachable();
#endif
    break;
  case GDAVendor::IONIC:
    ionic_initialize_gpu_qp(gpu_qp, conn_num);
    break;
  case GDAVendor::BNXT:
    bnxt_initialize_gpu_qp(gpu_qp, conn_num);
    break;
  case GDAVendor::MLX5:
    mlx5_initialize_gpu_qp(gpu_qp, conn_num);
    break;
```
Could do something like this and use `-Werror=switch-enum` to catch non-exhaustive switches - useful if we add other GDA vendors in the future. By the time we get here, `gda_vendor` should never be `NONE`.

src/gda/backend_gda.hpp
```diff
@@ -337,9 +331,14 @@ class GDABackend : public Backend {
...
 
   /**
-   * @brief Detect the available direct verbs libraries
+   * @brief Detect and load the available direct verbs libraries
    */
-  void autodetect_dv_libs();
+  void open_dv_libs();
+
+  /**
+   * @ brief Close opened direct versbs libraries
```
omor1:
```suggestion
   * @ brief Close opened direct verbs libraries
```

src/gda/backend_gda.hpp
```diff
@@ -536,6 +541,32 @@ class GDABackend : public Backend {
...
+  /**
+   * @brief open mlx5 dv lib
+   */
+  static void* mlx5_dv_dlopen();
+
+  /**
+   * @brief structures holding the function pointers to the direct verbs functionality
+   * of each network driver.
+   */
+  ionicdv_funcs_t ionic_dv;
```
omor1, line 536:
```suggestion
  extern ionicdv_funcs_t ionic_dv;
```
This is a header, globals should be extern. Define `ionic_dv` in `backend_gda.cpp`.

Actually, I think this can be local to `src/gda/ionic/backend_gda_ionic.cpp`, I don't think it's used anywhere else.
```C++
  static ionicdv_funcs_t ionic_dv;
```

src/gda/backend_gda.hpp
```diff
@@ -521,11 +521,16 @@ class GDABackend : public Backend {
...
+   * @brief open bnxt dv lib
+   */
+  static void* bnxt_dv_dlopen();
+
   /**
    * @brief structures holding the function pointers to the direct verbs functionality
    * of each network driver.
    */
-  mlx5dv_funcs_t mlx5dv_ftable_;
+  mlx5dv_funcs_t mlx5dv;
```
omor1, line 515: Define in `src/gda/mlx5/backend_gda_mlx5.cpp`, see comments about `ionic_dv`.

src/gda/backend_gda.hpp
```diff
@@ -509,7 +509,7 @@ class GDABackend : public Backend {
    * @brief structures holding the function pointers to the direct verbs functionality
    * of each network driver.
    */
-  bnxtdv_funcs_t bnxtdv_ftable_;
+  bnxtdv_funcs_t bnxt_re_dv;
```
omor1, line 494: `bnxt_re_dv` is used in `src/gda/backend_gda.cpp` only in `GDABackend::cleanup_ibv`, `GDABackend::modify_qps_init_to_rtr`, and `GDABackend::modify_qps_rtr_to_rts`. Can this functionality be moved to `src/gda/bnxt/backend_gda_bnxt.cpp`?

If so, `bnxt_re_dv` can be defined there (`static bnxtdv_funcs_t bnxt_re_dv;`) and this provider-specific detail doesn't need to be exposed. Even if it's easier to keep it here, it at least needs to be declared `extern` in the header and defined inside `src/gda/backend_gda.cpp`.

omor1: Ditto for `void *bnxtdv_handle_`.

src/gda/backend_gda.hpp
omor1: Ditto for `void *mlx5dv_handle_`.

src/gda/backend_gda.hpp
```diff
@@ -536,6 +541,32 @@ class GDABackend : public Backend {
...
+  /**
+   * @brief structures holding the function pointers to the direct verbs functionality
+   * of each network driver.
+   */
+  ionicdv_funcs_t ionic_dv;
+
+  /**
+   * @brief handle used for the dlopen of the IONIC library
+   */
+  void *ionicdv_handle_{nullptr};
```
omor1, line 541: Remove here and define as `static` in `src/gda/ionic/backend_gda_ionic.cpp`, or keep here as `extern`.

src/gda/bnxt/backend_gda_bnxt.cpp
```diff
@@ -214,33 +214,40 @@ void GDABackend::bnxt_create_qps(int sq_length) {
...
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, create_cq);
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, destroy_cq);
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, cq_mem_alloc);
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, umem_reg);
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, umem_dereg);
-  DLSYM_HELPER(bnxtdv_ftable_, bnxt_re_dv_, bnxtdv_handle_, get_default_db_region);
+int GDABackend::bnxt_dv_dl_init() {
+  bnxtdv_handle_ = bnxt_dv_dlopen();
+  if (!bnxtdv_handle_)
+    return ROCSHMEM_ERROR;
```
omor1, line 238:
```suggestion
  if (!bnxtdv_handle_) {
    return ROCSHMEM_ERROR;
  }
```

src/gda/backend_gda.cpp
```diff
@@ -673,13 +677,26 @@ void GDABackend::autodetect_dv_libs() {
...
 
+void GDABackend::close_dv_libs() {
+  if (ionicdv_handle_ != nullptr)
+    dlclose(ionicdv_handle_);
+
+  if (bnxtdv_handle_ != nullptr)
+    dlclose(bnxtdv_handle_);
+
+  if (mlx5dv_handle_ != nullptr)
+    dlclose(mlx5dv_handle_);
```
omor1, line 696:
```suggestion
  if (ionicdv_handle_) {
    dlclose(ionicdv_handle_);
  }

  if (bnxtdv_handle_) {
    dlclose(bnxtdv_handle_);
  }

  if (mlx5dv_handle_) {
    dlclose(mlx5dv_handle_);
  }
```

Comparison with `nullptr` is redundant, add braces to avoid future errors e.g. `goto fail`.

src/gda/ionic/backend_gda_ionic.cpp
```diff
@@ -0,0 +1,122 @@
...
+      DPRINTF("Could not open libionic.so. Returning\n");
+    }
+  }
+  return dv_handle;
+}
+
+int GDABackend::ionic_dv_dl_init() {
+  ionicdv_handle_ = ionic_dv_dlopen();
+  if (!ionicdv_handle_)
+    return ROCSHMEM_ERROR;
```
omor1, line 109:
```suggestion
  if (!ionicdv_handle_) {
    return ROCSHMEM_ERROR;
  }
```

src/gda/mlx5/backend_gda_mlx5.cpp
```diff
@@ -0,0 +1,122 @@
...
+  if (!dv_handle) {
+    DPRINTF("Could not open libmlx5.so. Returning\n");
+  }
+  return dv_handle;
+}
+
+int GDABackend::mlx5_dv_dl_init() {
+  mlx5dv_handle_ = mlx5_dv_dlopen();
+  if (!mlx5dv_handle_)
+    return ROCSHMEM_ERROR;
```
omor1, line 42:
```suggestion
  if (!mlx5dv_handle_) {
    return ROCSHMEM_ERROR;
  }
```

src/gda/ionic/queue_pair_ionic.cpp
```diff
@@ -0,0 +1,312 @@
...
+  /* Check if my cqe color == expected color */
+  uint32_t qtf_be = *(volatile uint32_t *)(&cqe->qid_type_flags);
+  if ((qtf_be & qtf_color_bit) != qtf_color_exp) {
+    return;
+  }
+
+  uint32_t msn = swap_endian_val<uint32_t>(cqe->send.msg_msn);
+
+  /* Report if the completion indicates an error. */
+  if (!!(qtf_be & swap_endian_val<uint32_t>(IONIC_V1_CQE_ERROR))) {
```
omor1, line 102: Why `!!`? That just turns any nonzero value into a 1 (or `true`), but since it's only being used in an `if` statement it's entire extraneous.

src/gda/ionic/queue_pair_ionic.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+  sq_msn = msn;
+}
+
+__device__ void QueuePair::ionic_quiet_internal(uint64_t activemask, uint32_t cons) {
+  uint32_t greed = 10;
+
+  /* wait for sq_msn to catch up or pass cons. */
+  /* 0x800000 - sign bit for 24-bit fields     */
+  while ((sq_msn - cons) & 0x800000) {
```
omor1, line 146: Define a constant for `0x800000`?
```C++
#define SIGN_BIT24 (0x800000)
static constexpr uint32_t sign_bit24{0x800000};
```

src/gda/ionic/queue_pair_ionic.cpp
```diff
@@ -0,0 +1,312 @@
...
+  if (((cq_pos - cq_dbpos) & cq_mask) >= 100) {
+    cq_dbpos = cq_pos;
+    __atomic_store_n(cq_dbreg, cq_dbval | (cq_mask & cq_dbpos), __ATOMIC_SEQ_CST); //TODO:maybe relaxed?
+  }
+
+  sq_msn = msn;
+}
+
+__device__ void QueuePair::ionic_quiet_internal(uint64_t activemask, uint32_t cons) {
+  uint32_t greed = 10;
```
omor1, line 142: What does `greed` mean? It seems to be a limit on the number of loop iterations that are executed below, but the variable name isn't intuitive.

src/gda/mlx5/queue_pair_mlx5.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+namespace rocshmem {
+
+__device__ void QueuePair::mlx5_ring_doorbell(uint64_t db_val, uint64_t my_sq_counter) {
+  swap_endian_store(const_cast<uint32_t*>(dbrec), (uint32_t)my_sq_counter);
+  __atomic_signal_fence(__ATOMIC_SEQ_CST);
+
+  __hip_atomic_store(db.ptr, db_val, __ATOMIC_SEQ_CST, __HIP_MEMORY_SCOPE_SYSTEM);
+  uint64_t db_uint = __hip_atomic_load(&db.uint, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+  db_uint ^= 0x100;
```
omor1, line 39: What is this constant?

src/gda/mlx5/queue_pair_mlx5.cpp
```diff
@@ -0,0 +1,273 @@
...
+  __hip_atomic_store(&db.uint, db_uint, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+}
+
+__device__ void QueuePair::mlx5_quiet() {
+  constexpr size_t BROADCAST_SIZE = 1024 / WF_SIZE;
```
omor1, line 44: Not an issue specifically here, but more broadly: I'd prefer to use a constant defined somewhere e.g. `constexpr size_t MAX_THREADS_PER_BLOCK{1024}` and `constexpr size_t MAX_WF_PER_BLOCK = MAX_THREADS_PER_BLOCK / WF_SIZE`. Then using e.g. `constexpr size_t BROADCAST_SIZE = MAX_WF_PER_BLOCK` makes it a lot clearer why the array is that size.

This is something that shows up in a couple places though and I don't think it makes sense to deal with that in this PR.

src/gda/queue_pair.cpp
```diff
@@ -66,22 +64,33 @@ QueuePair::QueuePair(struct ibv_pd* pd, int gda_vendor) {
...
+  case GDAVendor::MLX5:
     gda_op_rdma_write = MLX5_OPCODE_RDMA_WRITE;
     gda_op_rdma_read  = MLX5_OPCODE_RDMA_READ;
     gda_op_atomic_fa  = MLX5_OPCODE_ATOMIC_FA;
     gda_op_atomic_cs  = MLX5_OPCODE_ATOMIC_CS;
+    break;
+#endif //defined(GDA_MLX5)
+  default:
+    assert(false /* invalid nic provider */);
   }
```
omor1, line 94:
```diff
   switch (gda_vendor) {
+  case GDAVendor::NONE:
+    assert(false /* invalid nic provider */);
+#ifdef NDEBUG
+    __builtin_unreachable();
+#endif
+    break;
-#if defined(GDA_IONIC)
   case GDAVendor::IONIC:
     gda_op_rdma_write = IONIC_V2_OP_RDMA_WRITE;
     gda_op_rdma_read  = IONIC_V2_OP_RDMA_READ;
     gda_op_atomic_fa  = IONIC_V2_OP_ATOMIC_FA;
     gda_op_atomic_cs  = IONIC_V2_OP_ATOMIC_CS;
     break;
-#endif //defined(GDA_IONIC)
-#if defined(GDA_BNXT)
   case GDAVendor::BNXT:
     gda_op_rdma_write = BNXT_RE_WR_OPCD_RDMA_WRITE;
     gda_op_rdma_read  = BNXT_RE_WR_OPCD_RDMA_READ;
     gda_op_atomic_fa  = BNXT_RE_WR_OPCD_ATOMIC_FA;
     gda_op_atomic_cs  = BNXT_RE_WR_OPCD_ATOMIC_CS;
     break;
-#endif //defined(GDA_BNXT)
-#if defined(GDA_MLX5)
   case GDAVendor::MLX5:
     gda_op_rdma_write = MLX5_OPCODE_RDMA_WRITE;
     gda_op_rdma_read  = MLX5_OPCODE_RDMA_READ;
     gda_op_atomic_fa  = MLX5_OPCODE_ATOMIC_FA;
     gda_op_atomic_cs  = MLX5_OPCODE_ATOMIC_CS;
     break;
-#endif //defined(GDA_MLX5)
-  default:
-    assert(false /* invalid nic provider */);
   }
```
All vendors are compiled, so there's no reason for the `#if` checks. Explicitly indicate that `GDAVendor::NONE` is invalid here and use `-Werror=switch-enum` to ensure all cases are checked.

omor1 (commented):
Mostly lgtm, just some small nitpicks; they can mostly be addressed in subsequent PRs I think.

Is it necessary to still have compile-time vendor selection? I think compilation is now entirely self-contained and we don't need to link to any vendor-specific libraries or headers. Right now there is a mix of runtime and compile-time selection and I don't know if there are issues the occur when the provider for some vendor is not compiled.

src/gda/backend_gda.cpp
abouteiller: correct

src/gda/queue_pair.cpp
abouteiller: Preparing for making compilation optional again

src/gda/backend_gda.hpp
```diff
@@ -40,31 +40,23 @@
...
+#define DLSYM_HELPER(func_struct, prefix, handle, func_name)                                \
+do {                                                                                        \
+  *(void **) (&func_struct.func_name) = dlsym(handle, STRINGIFY(PPCAT(prefix, func_name))); \
+  if (!func_struct.func_name) {                                                             \
+    DPRINTF("Failed to find function %s \n",  STRINGIFY(PPCAT(prefix, func_name)));         \
+    dlclose(handle);                                                                        \
+    handle = nullptr;                                                                       \
+    return ROCSHMEM_ERROR;                                                                  \
+  }                                                                                         \
+} while (0)
```
edgargabriel: Can we not use the version of this macros that are in src/util.hpp ?

src/gda/backend_gda.cpp
```diff
@@ -599,22 +588,22 @@ void GDABackend::cleanup_ibv() {
...
-      err = bnxtdv_ftable_.destroy_qp(qps[i]);
+      err = bnxt_re_dv.destroy_qp(qps[i]);
       CHECK_ZERO(err, "bnxt_re_dv_destroy_qp");
 
-      err = bnxtdv_ftable_.umem_dereg(bnxt_qps[i].attr.rq_umem_handle);
+      err = bnxt_re_dv.umem_dereg(bnxt_qps[i].attr.rq_umem_handle);
       CHECK_ZERO(err, "bnxt_re_dv_umem_dereg (RQ)");
 
-      err = bnxtdv_ftable_.umem_dereg(bnxt_qps[i].attr.sq_umem_handle);
+      err = bnxt_re_dv.umem_dereg(bnxt_qps[i].attr.sq_umem_handle);
```
edgargabriel, line 597: not sure I am keen on renaming the function tables  from bnxty_dv_ftable to bnxt_re_dv (if for no other reason is that we also use the term ftable for function table for some other dlopen'd libraries (MPI and mlx5 is still using that nomenclature)). But keep it this way for now.

src/gda/backend_gda.cpp
```diff
@@ -673,13 +677,26 @@ void GDABackend::autodetect_dv_libs() {
       DPRINTF("Initializing rocSHMEM MLX5 GDA support failed\n");
     }
   }
+#endif // defined(GDA_MLX5)
 
   if (gda_vendor == GDAVendor::NONE) {
-    printf("Initializing rocSHMEM with IONIC, BNXT, or MLX5 GDA support failed\n");
+    DPRINTF("Initializing rocSHMEM with IONIC, BNXT, or MLX5 GDA support failed: no DV library found\n");
```
edgargabriel: since we call abort in the next line, I would prefer to get the error message always printed, not just if the user activates DEBUG statements. The user should know why the code aborted.

src/gda/queue_pair.cpp
edgargabriel: definitely leave the ```#if defined()``` statements in there. We still want to have the option to compile just for a specific vendor and avoid all switch statements.

src/gda/mlx5/queue_pair_mlx5.cpp
Yiltan: It flip flops between the two blueflame buffers. See table 13 in Section 7.2.2 of the Mellanox PRM

Yiltan (dismissed):

src/gda/backend_gda.cpp
abouteiller: Its a good idea for a cleanup pr later when we are less rushed

src/gda/backend_gda.hpp
abouteiller: Its not a global, its a private member of backend_gda class; suggesion to make it a local to the ionic/backend_gda_ionic would work for ionic but not for bnxt without further refactoring

src/gda/ionic/queue_pair_ionic.cpp
```diff
@@ -0,0 +1,312 @@
...
+  // exclude threads operating on a different qp from this thread lane mask
+  #pragma unroll
+  for (int i = 0; i < 64; ++i) {
+    uint64_t bit_i = 1ull << i;
+    if ((lane_mask & bit_i) && __shfl(this_val, i) != this_val) {
+      lane_mask &= ~bit_i;
+    }
+  }
+
+  return lane_mask;
```
avinashkethineedi, line 45:
```suggestion
  // Bitmask of lanes in this warp whose value == this_val
  uint64_t same_qp_mask = __match_any_sync(active, this_val);

  return same_qp_mask;
```

This for loop gets executed every time an RMA or atomic call is made. The `__match_any_sync` warp-level intrinsic achieves the same result as that loop

abouteiller: I do not want to introduce actual code change in a PR that rename/shuffles code around, but that is a good point for a followup.

Yiltan (approved):

avinashkethineedi (approved):

PR #273, opened 2025-10-07
Make ROCSHMEM_DISABLE_MIXED_IPC a synonym for ROCSHMEM_RO_DISABLE_IPC, ROCSHMEM_DISABLE_IPC

abouteiller:
## Motivation

We have communicated to downstream users that they should use ROCSHMEM_DISABLE_IPC as a control for the IPC feature. We inadvertently reverted that change with PR #235 

## Technical Details

* We introduce the new unified control ROCSHMEM_DISABLE_MIXED_IPC
* We check for both RO_DISABLE_IPC and DISABLE_IPC, and issue a deprecation warning if set
* In all cases, we set the ipcimpl.empty as the copy provider accordingly. 

## Test Plan

 * [x] GDA-IPC, GDA+IPC, RO+IPC, RO-IPC work as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,8 +121,8 @@ rocSHMEM has the following enviroment variables:
                         Defines the size of the rocSHMEM symmetric heap
                         Note the heap is on the GPU memory.
 
-    ROCSHMEM_RO_DISABLE_IPC (default : 0)
-                        Disables IPC support for the reverse offload backend.
+    ROCSHMEM_RO_DISABLE_MIXED_IPC (default : 0)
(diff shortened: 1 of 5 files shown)


edgargabriel:
DIdn't we agree that we will call it slightly different than ROCSHMEM_DISABLE_IPC, to avoid confusion with disabling IPC in the all-backend builds? This is about disabling IPC when used with RO or GDA simultaneously

omor1:
This works as a temporary workaround until I get proper alias support hooked up.

Yiltan (approved):

PR #274, opened 2025-10-08
Add host API for enqueuing barrier on given stream

drprajap:
## Motivation

Triton-dist has usecase to issue barrier on specific stream in case of ag + gemm and other such workloads when comp and comm are running on separate streams.

## Technical Details

This change adds host side api to launch a kernel on a stream to issue barrier_all device kernel.

## Test Plan

Tested as part of triton-dist integration

## Submission Checklist

### DIFF
--- a/include/rocshmem/rocshmem_COLL.hpp
+++ b/include/rocshmem/rocshmem_COLL.hpp
@@ -599,6 +599,12 @@ __host__ int rocshmem_ctx_double_prod_reduce(
     rocshmem_ctx_t ctx, rocshmem_team_t team, double *dest, const double *source,
     int nreduce);
 
+/**
+ * @brief kernel for performing a barrier synchronization.
+ * Caller enqueues the kernel on given stream
+ *
+ * @return void
+ */
 __global__ ATTR_NO_INLINE void rocshmem_barrier_all_kernel();
 
 /**
--- a/include/rocshmem/rocshmem_COLL.hpp
+++ b/include/rocshmem/rocshmem_COLL.hpp
@@ -599,6 +599,8 @@ __host__ int rocshmem_ctx_double_prod_reduce(
     rocshmem_ctx_t ctx, rocshmem_team_t team, double *dest, const double *source,
     int nreduce);
 
+__global__ ATTR_NO_INLINE void rocshmem_barrier_all_kernel();
+
 /**
  * @brief perform a collective barrier between all PEs in the system.
  * The caller is blocked until the barrier is resolved.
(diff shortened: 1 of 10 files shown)


include/rocshmem/rocshmem_COLL.hpp
```diff
@@ -599,6 +599,14 @@ __host__ int rocshmem_ctx_double_prod_reduce(
     rocshmem_ctx_t ctx, rocshmem_team_t team, double *dest, const double *source,
     int nreduce);
 
+/**
+ * @brief kernel for performing a barrier synchronization.
+ * Caller enqueues the kernel on given stream
+ *
+ * @return void
+ */
+__global__ ATTR_NO_INLINE void rocshmem_barrier_all_kernel();
```
Yiltan, line 608: This exposes `rocshmem_barrier_all_kernel` to users. Is this something Triton requires?

Or does Triton only require the `rocshmem_barrier_all_on_stream` API?

drprajap: Triton only requires rocshmem_barrier_all_on_stream API, so kernel function can be moved to somewhere internal I believe, can you suggest which is better place not to expose this to users?

wenlei-bao:
Looking forward to this change!

Yiltan (approved):

avinashkethineedi:
This works only on default context

drprajap:
> This works only on default context

wouldn't this ->  https://github.com/ROCm/rocSHMEM/blob/4580a958aa14d4ac47da94306da917386d98357c/src/context_host.cpp#L121 take care of calling into correct context class based on Use_IPC flag?

This API should work same way as rocshmem_barrier_all which also uses ROCSHMEM_HOST_CTX_DEFAULT ([Here](https://github.com/ROCm/rocSHMEM/blob/4580a958aa14d4ac47da94306da917386d98357c/src/rocshmem.cpp#L882)) to call into context specific definition, so I followed the same implementation, Can you elaborate if I misunderstood something?

avinashkethineedi:
> > This works only on default context
> 
> wouldn't this ->
> 
> https://github.com/ROCm/rocSHMEM/blob/4580a958aa14d4ac47da94306da917386d98357c/src/context_host.cpp#L121
> 
> take care of calling into correct context class based on Use_IPC flag?
> This API should work same way as rocshmem_barrier_all which also uses ROCSHMEM_HOST_CTX_DEFAULT ([Here](https://github.com/ROCm/rocSHMEM/blob/4580a958aa14d4ac47da94306da917386d98357c/src/rocshmem.cpp#L882)) to call into context specific definition, so I followed the same implementation, Can you elaborate if I misunderstood something?

Yes, it should. I was looking into the device-side implementation, but the context-specific barriers use a different set of APIs than what this PR modifies. The PR looks fine for now

avinashkethineedi (approved):

PR #279, opened 2025-10-16
[GDA/BNXT] Implemented CQE Collapsing 

Yiltan:
## Motivation

Use Broadcom's CQE Collapsing Feature

## Technical Details

- Simplified CQ logic
  - No longer required to track phase or ring doorbell
  - CQE is always valid
  - Use cons_idx to work out what free slots are available in SQ
- Requires a CQ for both SQ and RQ. They are denoted as SCQ and RCQ.
- Removed quiet from `post_wqe_rma`
- Check that we have the correct firmware version

## Test Plan

- [x] Validate DeepEP
- [x] Validate Functional Tests


Yiltan:
Updated firmware comparison, rebased with develop, re-tested with deepep

Yiltan:
Please wait for review, #280 updated the `gda_vendor` variable this PR needs to be reworked.

Yiltan:
Rebased and re-tested. This is good to review

abouteiller (approved):

PR #280, opened 2025-10-16
Add ROCSHMEM_GDA_PROVIDER envvar 

abouteiller:
## Motivation

Let user force-control the selection logic for GDA providers

## Technical Details

* Added ROCSHMEM_GDA_PROVIDER envvar, should contain a single word from list {ionic,bnxt,mlx5}
* renamed GDAVendor to GDAProvider to match the envvar

Grafted on top of #272, so will need rebased

## Test Plan

* [x] runs and pick the right provider when not set 
* [x] runs and pick the selected provider when set (GDA backend does not initialize when selecting ionic on bnxt hw, abort after pd_create when selecting mlx5 on bnxt hw, both are incorrect selection so expected outcome)

### DIFF
--- a/src/envvar.cpp
+++ b/src/envvar.cpp
@@ -65,6 +65,7 @@ namespace envvar {
   }  // namespace ro
 
   namespace gda {
+    const var<std::string> provider("PROVIDER", "");
     const var<bool> alternate_qp_ports("ALTERNATE_QP_PORTS", "", true);
     const var<uint8_t> traffic_class("TRAFFIC_CLASS", "", 0);
   }  // namespace gda
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -528,23 +528,23 @@ void GDABackend::rte_barrier() {
   }
 }
 
-GDAVendor GDABackend::requested_provider() {
+GDAProvider GDABackend::requested_provider() {
   /* Check whether the user explicitely requests a particular provider type */
   std::string envstr = envvar::gda::provider;
   std::transform(envstr.begin(), envstr.end(), envstr.begin(), ::tolower);
   if (!envstr.empty()) {
     printf("Found environment variable ROCSHMEM_GDA_PROVIDER, value is %s\n", envstr.c_str());
     if (envstr.find("bnxt") != std::string::npos) {
-      return GDAVendor::BNXT;
+      return GDAProvider::BNXT;
     }
     if (envstr.find("ionic") != std::string::npos) {
-      return GDAVendor::IONIC;
+      return GDAProvider::IONIC;
     }
     if (envstr.find("mlx5") != std::string::npos) {
-      return GDAVendor::MLX5;
+      return GDAProvider::MLX5;
     }
   }
-  return GDAVendor::NONE;
+  return GDAProvider::UNSET;
 }
 
 /* Currently we only check whether we can dlopen a Direct Verbs library.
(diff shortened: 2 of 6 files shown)


src/gda/backend_gda.cpp
```diff
@@ -528,37 +528,63 @@ void GDABackend::rte_barrier() {
...
+GDAProvider GDABackend::requested_provider() {
+  /* Check whether the user explicitely requests a particular provider type */
+  std::string envstr = envvar::gda::provider;
+  std::transform(envstr.begin(), envstr.end(), envstr.begin(), ::tolower);
+  if (!envstr.empty()) {
+    printf("Found environment variable ROCSHMEM_GDA_PROVIDER, value is %s\n", envstr.c_str());
+    if (envstr.find("bnxt") != std::string::npos) {
+      return GDAProvider::BNXT;
+    }
+    if (envstr.find("ionic") != std::string::npos) {
```
edgargabriel, line 540: this is maybe up for debate, the public name of the NIC is not ionic, its ainic, ionic is the historic name of the driver

edgargabriel (approved):

src/gda/backend_gda.cpp
```diff
@@ -636,50 +662,54 @@ void GDABackend::cleanup_ibv() {
...
 
-  //TODO: environment variable selection/deselection
   //this hardcoded init order will always prefer BNXT>IONIC>MLX5
-  //if all three drivers are installed
+  //if all three drivers are installed and enabled
 
 #if defined(GDA_BNXT)
-  if (gda_vendor == GDAVendor::NONE) {
+  if (gda_provider == GDAProvider::UNSET
+  && (requested == GDAProvider::UNSET || requested == GDAProvider::BNXT)) {
```
Yiltan, line 675: Is `A && (A || B)` the same as `A`?  I'm trying to understand the logical expression

Screenshot 2025-10-17 at 9 03 21 AM


I assume we want this, no?

```
if  (requested == GDAProvider::UNSET || requested == GDAProvider::BNXT) {
```

abouteiller: first test is on gda_provider (already initialized another provider), second test is on requested (do I want a specific provider)

Yiltan: Ah, I missed that

Yiltan (approved):

src/envvar.cpp
```diff
@@ -65,6 +65,7 @@ namespace envvar {
   }  // namespace ro
 
   namespace gda {
+    const var<std::string> provider("PROVIDER", "");
```
omor1, line 68: I'm a bit late on noticing this, but this should really follow the examples for e.g. `socket_family` or `debug_level`, so that doesn't need to be parsed inside `GDABackend::requested_provider()`.

```C++
// src/envvar.hpp
using var_types = unique_type_sequence_t<bool,
                                         /* ... */,
                                         types::socket_family,
                                         types::debug_level,
                                         GDAProvider>;

namespace types {
  inline namespace _gda {
    std::istream& operator>>(std::istream& is, GDAProvider& provider);
    std::ostream& operator<<(std::ostream& os, const GDAProvider& provider);
  }  // inline namespace _gda
}  // namespace types

namespace gda {
  extern const var<GDAProvider> provider;
}  // namespace gda



// src/envvar.cpp
namespace gda {
  const var<GDAProvider> provider("PROVIDER", "", GDAProvider::UNSET);
}  // namespace gda

namespace types {
  inline namespace _gda {
    std::istream& operator>>(std::istream& is, GDAProvider& provider) {
      std::string provider_str;
      is >> provider_str;
      if (provider_str == "ionic" ||
          provider_str == "pensando") {
        provider = GDAProvider::IONIC;
      } else if (provider_str == "bnxt") {
        provider = GDAProvider::BNXT;
      } else if (provider_str == "mlx5") {
        provider = GDAProvider::MLX5;
      } else {
        // all other inputs are invalid
        is.setstate(std::ios_base::failbit);
        family = GDAProvider::UNSET;
      }
      return is;
    }

    std::ostream& operator<<(std::ostream& os, const GDAProvider& provider) {
      switch (provider) {
      case GDAProvider::UNSET:
        return os << "unset";
      case GDAProvider::IONIC:
        return os << "ionic";
      case GDAProvider::BNXT:
        return os << "bnxt";
      case GDAProvider::MLX5:
        return os << "mlx5";
      }
    }
  }  // inline namespace _gda
}  // namespace types
```

PR #281, opened 2025-10-17
update tester for RO

edgargabriel:
## Motivation

update the tester script to only tests the amo functions on RO that are expected to pass. We can revisit the non-passing tests later, but this prevents us from having passing CIs at the moment, while RO is simply lower priority than other asks.

## Submission Checklist

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -289,7 +289,7 @@ TestRMA() {
   TestRMAGet
 }
 
-TestAMO() {
+TestAMORO() {
   ##############################################################################
   #       | Name             | Ranks | Workgroups | Threads | Max Message Size #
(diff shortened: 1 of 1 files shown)


abouteiller (approved):

abouteiller:
Looks like fetch, signal_fetch and others are also failing

edgargabriel:
> Looks like fetch, signal_fetch and others are also failing

they are failing on the gfx942 node, but on all other systems they are passing.

edgargabriel:
Can confirm based on the CI, the gfx942 tests are failing, but on all other systems (gfx950, gfx90a with all three OS variants tested) the CI passes. So we will have to figure out the gfx942 issue, but this PR resolved the general issue with the tests, the rest is a platform specific problem.

PR #282, opened 2025-10-17
Implement rocshmem_pe_quiet()

Yiltan:
rocSHMEM does not have an equivalent to `shmem_pe_quiet`

### DIFF
--- a/docs/api/memory_ordering.rst
+++ b/docs/api/memory_ordering.rst
@@ -38,7 +38,7 @@ This routine completes all previous operations posted to this context.
 ROCSHMEM_PE_QUIET
 --------------
 
-.. cpp:function:: __device__ void rocshmem_ctx_quiet(shmem_ctx_t ctx, const int *target_pes, size_t npes)
+.. cpp:function:: __device__ void rocshmem_ctx_pe_quiet(shmem_ctx_t ctx, const int *target_pes, size_t npes)
 .. cpp:function:: __device__ void rocshmem_quiet(const int *target_pes, size_t npes)
 
  :param ctx: Context with which to perform this operation.
--- a/docs/api/memory_ordering.rst
+++ b/docs/api/memory_ordering.rst
@@ -39,7 +39,7 @@ ROCSHMEM_PE_QUIET
 --------------
 
 .. cpp:function:: __device__ void rocshmem_ctx_pe_quiet(shmem_ctx_t ctx, const int *target_pes, size_t npes)
-.. cpp:function:: __device__ void rocshmem_quiet(const int *target_pes, size_t npes)
+.. cpp:function:: __device__ void rocshmem_pe_quiet(const int *target_pes, size_t npes)
 
  :param ctx: Context with which to perform this operation.
  :param target_pes: Address of target PE array where the operations need to be completed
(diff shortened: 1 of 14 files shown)


docs/api/memory_ordering.rst
```diff
@@ -34,3 +34,18 @@ ROCSHMEM_QUIET
 
 **Description:**
 This routine completes all previous operations posted to this context.
+
+ROCSHMEM_PE_QUIET
+--------------
+
+.. cpp:function:: __device__ void rocshmem_ctx_quiet(shmem_ctx_t ctx, const int *target_pes, size_t npes)
```
abouteiller:
```suggestion
.. cpp:function:: __device__ void rocshmem_ctx_pe_quiet(shmem_ctx_t ctx, const int *target_pes, size_t npes)
```

docs/api/memory_ordering.rst
```diff
@@ -34,3 +34,18 @@ ROCSHMEM_QUIET
...
+.. cpp:function:: __device__ void rocshmem_quiet(const int *target_pes, size_t npes)
```
abouteiller:
```suggestion
.. cpp:function:: __device__ void rocshmem_pe_quiet(const int *target_pes, size_t npes)
```

abouteiller (changes requested):

abouteiller (dismissed):

src/ipc/context_ipc_device.cpp
```diff
@@ -96,6 +96,12 @@ __device__ void IPCContext::quiet() {
   fence();
 }
 
+__device__ void IPCContext::pe_quiet(const int *target_pes, size_t npes) {
+  for (int i = 0; i < npes; i++) {
+    fence(target_pes[i]);
```
avinashkethineedi: Context can be tied to a team, the PEs in the array needs to be converted to PE numbering relative to the world size.

src/rocshmem_gpu.cpp
```diff
@@ -485,6 +489,12 @@ __device__ void rocshmem_ctx_quiet(rocshmem_ctx_t ctx) {
   get_internal_ctx(ctx)->quiet();
 }
 
+__device__ void rocshmem_ctx_pe_quiet(rocshmem_ctx_t ctx, const int *target_pes, size_t npes) {
+  GPU_DPRINTF("Function: %s (ctx=%zd)\n", __FUNC__, ctx.ctx_opaque);
+
```
avinashkethineedi, line 494:
```suggestion
  for (int i = {0}; i < npes;  i++) {
    target_pes[i] = translate_pe(ctx, target_pes[i]);
  }
```

Yiltan:
resolved conflicts with develop

edgargabriel (dismissed):

avinashkethineedi (changes requested):

src/rocshmem_gpu.cpp
Yiltan: Good catch, this exact approach doesn't work as I got the compiler error `error: read-only variable is not assignable` when trying this.

I _think_ the way I did it still covers the case that you mentioned

avinashkethineedi (approved):

PR #285, opened 2025-10-20
add support for GPUs using wavefront size of 32

edgargabriel:
## Motivation

This Pr introduces support for wavefront size of 32.
In order to test the PR, it also adds experimental support for gfx1100 (Radeon RX7900 GPUs, and Radeon Pro W7900) GPUs.

## Test Plan

- unit tests are passing
- functional tests are mostly passing, with few exceptions:
```
TESTS FAILED:  teamreduction_n2_w1_z1_32768B pingpong_n2_w1_z1 pingpong_n2_w8_z1 pingpong_n2_w32_z1 pingall_n2_w1_z1 pingall_n2_w8_z1 pingall_n2_w32_z1
```
Investigating the failures, but wanted to provide a first preview of the PR. 
## Submission Checklist

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -117,6 +117,7 @@ ExecTest() {
   NUM_WG=$3
   NUM_THREADS=$4
   MAX_MSG_SIZE=$5
+  NUM_GPUS=$(amd-smi list | grep GPU | wc -l)
 
   TIMEOUT=$((5 * 60)) # Timeout in seconds
 
@@ -159,9 +160,13 @@ ExecTest() {
   CMD+=" >> $LOG_DIR/$TEST_LOG_NAME.log 2>&1"
 
   # Run Test
-  echo $TEST_LOG_NAME
-  echo "# $CMD" >"$LOG_DIR/$TEST_LOG_NAME.log"
-  eval $CMD
+  if [ $NUM_RANKS -le $NUM_GPUS ]; then
+    echo $TEST_LOG_NAME
+    echo "# $CMD" >"$LOG_DIR/$TEST_LOG_NAME.log"
+    eval $CMD
+  else
+    echo "Skipping test $TEST_LOG_NAME"
+  fi
 
   # Validate Test
   if [ $? -ne 0 ]
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -117,8 +117,16 @@ ExecTest() {
   NUM_WG=$3
   NUM_THREADS=$4
   MAX_MSG_SIZE=$5
-  NUM_GPUS=$(amd-smi list | grep GPU | wc -l)
 
+  if command -v amd-smi >/dev/null
+  then
+      NUM_GPUS=$(amd-smi list | grep GPU | wc -l)
+  elif command -v rocm-smi >/dev/null
+  then
+      NUM_GPUS=$(rocm-smi --showserial | grep GPU | wc -l)
+  else
+      NUM_GPUS=64
+  fi
   TIMEOUT=$((5 * 60)) # Timeout in seconds
 
   TEST_NUM=${TEST_NUMBERS[$TEST_NAME]}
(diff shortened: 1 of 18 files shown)


scripts/functional_tests/driver.sh
```diff
@@ -117,6 +117,7 @@ ExecTest() {
   NUM_WG=$3
   NUM_THREADS=$4
   MAX_MSG_SIZE=$5
+  NUM_GPUS=$(amd-smi list | grep GPU | wc -l)
```
abouteiller: I have many systems where this executable is not installed

edgargabriel: good point, starting from ROCm 7.0 it supposed be the default and from 7.2 on its a complete replacement for rocm-smi. Let me think how to handle. I just wanted to have a mechanism to skip tests that would need more GPUs than are available on the system.

scripts/functional_tests/driver.sh
```diff
@@ -159,9 +160,13 @@ ExecTest() {
   CMD+=" >> $LOG_DIR/$TEST_LOG_NAME.log 2>&1"
 
   # Run Test
-  echo $TEST_LOG_NAME
-  echo "# $CMD" >"$LOG_DIR/$TEST_LOG_NAME.log"
-  eval $CMD
+  if [ $NUM_RANKS -le $NUM_GPUS ]; then
```
Yiltan: I would also check if there is a valid `$HOSTFILE` here too.

If we run on two nodes, then num_ranks>num_gpus

src/gda/queue_pair.cpp
```diff
@@ -59,7 +59,11 @@ QueuePair::QueuePair(struct ibv_pd* pd, int gda_provider) {
     fetching_atomic_lkey = mr_fetching_atomic->lkey;
   }
 
-  for(int i{0}; i < FETCHING_ATOMIC_CNT; i+=WF_SIZE) {
+  int deviceId;
+  hipDeviceProp_t prop;
+  CHECK_HIP(hipGetDevice(&deviceId));
+  CHECK_HIP(hipGetDeviceProperties(&prop, deviceId));
```
Yiltan: We also call this API here: https://github.com/ROCm/rocSHMEM/blob/b534423de782fc35393dbd326729c927bb0542a0/src/reverse_offload/backend_ro.cpp#L62

Maybe it could be done in a central location?

scripts/functional_tests/driver.sh
abouteiller: `amd-smi list || rocm-smi  --showserial | grep GPU | wc -l`

edgargabriel:
units tests on gfx90a are failing, will have to look into this why that would be the case.

Yiltan (approved):

edgargabriel:
Can confirm that with the current version of the PR, all unit tests are passing on a system with GPUs that have wavefront size of 32, and all functional tests (except for the ones listed above, which fail for different reasons that have nothing to do with the wavefront size)

edgargabriel:
One final updated: setting `HSA_FORCE_FINE_GRAIN_PCIE=1` makes most of the failing tests pass, all variations of the pingpong and pingall tests are passing when setting this environment variable. This environment variable is required to enable fine-grained memory on PCIe connected GPUs. 

The only failing test on gfx1100 is the reduction tests with this setting.

abouteiller (approved):

abouteiller:
No changes needed for the build scripts? I thought you had to enable HDP_FLUSH.

edgargabriel:
> No changes needed for the build scripts? I thought you had to enable HDP_FLUSH.

no, I tried enabling HDP_FLUSH, but it didn't help on the Radeon systems, I think they don't have an HDP cache. So right now lets leave it as it is.

scripts/functional_tests/driver.sh
```diff
@@ -159,9 +168,13 @@ ExecTest() {
   CMD+=" >> $LOG_DIR/$TEST_LOG_NAME.log 2>&1"
 
   # Run Test
-  echo $TEST_LOG_NAME
-  echo "# $CMD" >"$LOG_DIR/$TEST_LOG_NAME.log"
-  eval $CMD
+  if [ $NUM_RANKS -le $NUM_GPUS ] && [[ "" == "$HOSTFILE" ]]; then
```
Yiltan, line 171: This seems to break multi-node testing, I'll create another PR shortly

PR #286, opened 2025-10-20
Provide an error when there are no NICs on a system

Yiltan:
## Motivation

Give users a better error message

### DIFF
--- a/src/gda/topology.cpp
+++ b/src/gda/topology.cpp
@@ -386,8 +386,9 @@ namespace rocshmem
       // Query the number of IBV devices
       int numIbvDevices = 0;
       ibv_device** deviceList = ibv_get_device_list(&numIbvDevices);
+      CHECK_NNULL(deviceList, "ibv_get_device_list");
 
-      if (deviceList && numIbvDevices > 0) {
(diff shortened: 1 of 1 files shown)


src/gda/topology.cpp
```diff
@@ -386,8 +386,9 @@ namespace rocshmem
       // Query the number of IBV devices
       int numIbvDevices = 0;
       ibv_device** deviceList = ibv_get_device_list(&numIbvDevices);
+      CHECK_NNULL(deviceList, "ibv_get_device_list");
```
abouteiller, line 389: Is it not going to abort here instead of printing the error mesage line 458?

abouteiller (approved):

src/gda/topology.cpp
Yiltan: Not nessiarly, from the man page:

> ibv_get_device_list() returns the array of available RDMA devices, or sets errno and returns NULL if the request fails. If no devices are found then num_devices is set to 0, and non-NULL is returned.

Checking null and the num_devices is checking two different situations

edgargabriel (approved):

PR #290, opened 2025-10-22
RFC - rocSHMEM Plotter

Yiltan:
## Motivation

Users want a tool compare performance between version X and version Y of our code.

## Design

A python matplotlib script that can be used to compare the log files of our functional tests

We would use this simple command line:

```
python3 rocshmem_plot.py <TestName> <Test1_Label> <Test1_Log_dir> ...  <TestN_Label> <TestN_Log_dir>
```

## Rendered README file

https://github.com/ROCm/rocSHMEM/blob/5eec94d501dc1ea49c5b3c313bb3f79f18361f82/tools/functional_test_plotting/README.md

## Example

example


gaoikawa:
@Yiltan - Enhancement requests (if you agree) - in a past life doing performance reporting/charting like this I found it useful to create a PDF of all the images, then it's easier to scroll through all the charts.

Alternatively, something interactive is always nice too, but then requires something like https://plotly.com/python/time-series/.  If you want the user to easily hover over a point in the line to get the value.

ammallya:

PR #292, opened 2025-10-24
[GDA] Improve error messages

Yiltan:
## Motivation

Currently, if a NIC has the incorrect firmware then rocSHMEM outputs the wrong firmware version.

If you run rocSHMEM with a large amount of PEs/nodes, then its difficult to work out which NIC is the issue.

This PR prints out the hostname and the NIC that is giving us problems. This makes it easier to work out which NIC has issues.

### DIFF
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -896,17 +896,17 @@ void GDABackend::validate_ib_device() {
 
     if (bnxt_vendor_id != device_attr.vendor_id) {
       printf("%s GDAProvider::BNXT requested but an invalid device is selected\n", debug_str.c_str());
-      abort();
+      exit(1);
     }
 
     if (supported_bnxt_part_ids.find(device_attr.vendor_part_id) == supported_bnxt_part_ids.end()) {
       printf("%s Unsupported Broadcom Part: %x\n", debug_str.c_str(), device_attr.vendor_part_id);
-      abort();
+      exit(1);
     }
 
     if (strverscmp(min_supported_bnxt_fw_ver, device_attr.fw_ver) > 0) {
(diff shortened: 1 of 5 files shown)


src/gda/backend_gda.cpp
```diff
@@ -885,19 +894,18 @@ void GDABackend::validate_ib_device() {
     const std::set<uint32_t> supported_bnxt_part_ids = { 0x1760 /* BCM57608 */};
     const char min_supported_bnxt_fw_ver[12] = "233.2.104.0";
 
-
     if (bnxt_vendor_id != device_attr.vendor_id) {
-      printf("GDAProvider::BNXT requested but an invalid device is selected\n");
+      printf("%s GDAProvider::BNXT requested but an invalid device is selected\n", debug_str.c_str());
       abort();
```
abouteiller: should these abort just be exit(1), since this is not a bug but a misconfigured system

abouteiller (dismissed):

src/gda/bnxt/backend_gda_bnxt.cpp
```diff
@@ -87,6 +87,9 @@ void GDABackend::bnxt_initialize_gpu_qp(QueuePair* gpu_qp, int conn_num) {
 
   /* Export Inline Threshold */
   gpu_qp->inline_threshold = inline_threshold;
+
+  /* Export Debug String*/
+  strcpy(gpu_qp->debug_str, debug_str.c_str());
```
abouteiller: strncpy

Yiltan: There seems to be an issue on the device-side debug statements. I've left only the host-side debug in this PR.

I can resolve the device-side debug later

edgargabriel (approved):

PR #294, opened 2025-10-27
minor change to MPI detection logic

edgargabriel:
## Motivation

somehow the test whether we requested MPI support or not stopped working, although no obvious code change can be located.

## Technical Details

Make the if-statement more stringent by explicitely testing whether USE_MPI_SUPPORT is "ON".

## Submission Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -168,7 +168,7 @@ if (NOT BUILD_TESTS_ONLY)
     set(BUILD_UNIT_TESTS OFF)
   endif()
 
-  if (USE_EXTERNAL_MPI)
+  if (USE_EXTERNAL_MPI STREQUAL "ON")
     if(NOT HAVE_EXTERNAL_MPI)
       message(FATAL_ERROR "External MPI support requested but MPI support not found. Build Aborted")
     endif()


CMakeLists.txt
avinashkethineedi, line 157: doesn't this do the same thing as in line [171](https://github.com/ROCm/rocSHMEM/pull/294/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR171)

edgargabriel: no, USE_EXTERNAL_MPI is the user request for MPI (which can be ON, OFF, or UNSET / AUTO). HAVE_EXTERNAL_MPI is whether we found external MPI support. That can be independent of what the user asked us to do (e.g. they did not explicitly ask for MPI support but also didn't say that they don't want MPI, but we might have found an MPI library)

gaoikawa (approved):

avinashkethineedi (approved):

CMakeLists.txt
```diff
@@ -168,7 +168,7 @@ if (NOT BUILD_TESTS_ONLY)
     set(BUILD_UNIT_TESTS OFF)
   endif()
 
-  if (USE_EXTERNAL_MPI)
+  if (USE_EXTERNAL_MPI STREQUAL "ON")
```
omor1, line 171: My preference for this check would be:
```suggestion
  if (USE_EXTERNAL_MPI AND NOT USE_EXTERNAL_MPI STREQUAL "AUTO")
```

In general:
```cmake
if (NOT USE_EXTERNAL_MPI)
  # USE_EXTERNAL_MPI is a false constant
endif()

if (USE_EXTERNAL_MPI STREQUAL "AUTO")
  # USE_EXTERNAL_MPI is AUTO
endif()

if (USE_EXTERNAL_MPI)
  # USE_EXTERNAL_MPI is not a false constant
  # i.e. either a true constant or AUTO
endif()

if (USE_EXTERNAL_MPI AND NOT USE_EXTERNAL_MPI STREQUAL "AUTO")
  # USE_EXTERNAL_MPI is neither a false constant nor AUTO
  # i.e. it is a true constant
  # if this an especially common check, can add a local variable e.g.:
  set(USE_EXTERNAL_MPI_ON TRUE)
else()
  # USE_EXTERNAL_MPI_ON is false when USE_EXTERNAL_MPI is AUTO or a false constant
  set(USE_EXTERNAL_MPI_ON FALSE)
endif()
```

edgargabriel: @omor1 question: does this line here not restrict the potential values of `USE_EXTERNAL_MPI` to `ON, OFF, and AUTO` 

https://github.com/ROCm/rocSHMEM/blob/develop/CMakeLists.txt#L77

If not, what is the precise meaning of that line?

PR #298, opened 2025-10-28
functional_tests: loop count option, disable invalid atomics

ahubbe-amd:
## Motivation

The disabled atomic tests fail assert of size.

## Technical Details

Disable atomic invalid sizes.

## Submission Checklist

- [x] Run atomic functional tests, observe normal exit without assert.


abouteiller:
This disables testing for other backends like IPC and RO, which is the reason we didn't already roll that solution.

PR #299, opened 2025-10-29
[GDA] Implement internal_direct_barrier_wg

Yiltan:
## Motivation

- If we know an entire workgroup is calling into a barrier operation, then we can make some optimizations

## Technical Details

- Each wave:
  - Waits for its own flag
  - issues its own puts
  - quiets its own PEs

## Test Plan

- Can be tested with alltoall collective

```
mpirun -n 32   --display-map -mca pml ucx -mca osc ucx -x ROCSHMEM_MAX_NUM_CONTEXTS=1 --map-by numa --timeout 300 --hostfile /home/ytemucin/hostfile.txt
 ./tests/functional_tests/rocshmem_functional_tests -a 19 -w 1 -z 1024 -s 1024
```

## Test Result

- If we launch the barrier with a single thread, there is little to no change in pefromance
- If we use the entire workgroup (`-z 1024`) then we see up-to 2x performance improvement in alltoall_wg

### DIFF
--- a/src/gda/context_gda_device_coll.cpp
+++ b/src/gda/context_gda_device_coll.cpp
@@ -37,6 +37,9 @@ __device__ void GDAContext::internal_direct_barrier(int pe, int PE_start,
   if (pe == PE_start) {
     // Go through all PE offsets (except current offset = 0)
     // and wait until they all reach
+#if defined(__gfx90a__)
+    __threadfence_system();
+#endif /* __gfx90a__ */
     for (int i = 1; i < n_pes; i++) {
       wait_until(&pSync[i], ROCSHMEM_CMP_EQ, flag_val);
       pSync[i] = ROCSHMEM_SYNC_VALUE;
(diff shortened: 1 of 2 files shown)


src/gda/context_gda_device_coll.cpp
```diff
@@ -70,6 +70,57 @@ __device__ void GDAContext::internal_direct_barrier(int pe, int PE_start,
...
+  if (pe == PE_start) {
+    int wf_id = get_flat_block_id() / WF_SIZE;
+    int wf_count = (int) ceil((double)get_flat_block_size() / (double)WF_SIZE);
+    bool wf_leader = 0 == get_active_lane_num();
+
+    // Go through all PE offsets (except current offset = 0)
+    // and wait until they all reach
+    if (wf_leader) {
+      for (int j = wf_id + 1; j < n_pes; j+= wf_count) {
+        wait_until(&pSync[j], ROCSHMEM_CMP_EQ, flag_val);
```
abouteiller, line 87: why not wait_until_all? is there a performance advantage doing this way?

abouteiller (approved):

src/gda/context_gda_device_coll.cpp
Yiltan: `wait_until_all` would serialize this portion so a single thread would loop over the psync buffer.

I compared the two solutions, there around 6% decrease in performance if we use `wait_until_all`  

Code modification I tested:

```
if (is_thread_zero_in_block()) {
  int status[1024];
  wait_until_all(&pSync[1], n_pes - 1, (const int*) &status, ROCSHMEM_CMP_EQ, flag_val);
  for (int j = 1; j < n_pes; j++) {
    pSync[j] = ROCSHMEM_SYNC_VALUE;
  }
}
```

PR #302, opened 2025-10-30
install_dependencies pip issues with ubuntu 24

abouteiller:
## Motivation

The proper way of using pip varies depends on what base OS we are on (e.g., what worked on Ubuntu 22 does not anymore on Ubuntu 24)

Make the INSTALL_DIR user selectable. **Prior behavior is preserved** if variable not set . 

## Technical Details

`python3 -m venv --system-site-packages --without-pip venv` works even if ubuntu package `ensurepip` is not installed. 

## Test Plan

Ran `intall_dependencies.sh` followed by `build_scripts/gda_bnxt` on both Ubuntu 22 and Ubuntu 24

### DIFF
--- a/scripts/install_dependencies.sh
+++ b/scripts/install_dependencies.sh
@@ -72,8 +72,9 @@ git clone --recursive $_OMPI_REPO
 cd ompi
 git checkout $_OMPI_COMMIT_HASH
 git submodule update --init --recursive
-python3 -m venv --system-site-packages venv
-venv/bin/python3 -m pip install -r docs/requirements.txt
+python3 -m venv --system-site-packages --without-pip venv
+. venv/bin/activate
+python3 -m pip install -r docs/requirements.txt
 ./autogen.pl
 ./configure --prefix=$_OMPI_INSTALL_DIR  \
             --with-rocm=$_ROCM_DIR       \
--- a/scripts/install_dependencies.sh
+++ b/scripts/install_dependencies.sh
@@ -32,19 +32,19 @@ if [[ -z "${_ROCM_DIR}" ]]; then
 fi
 
 # Location of dependencies source code
-BUILD_DIR=${BUILD_DIR:-$PWD}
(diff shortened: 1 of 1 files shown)


edgargabriel (approved):

scripts/install_dependencies.sh
```diff
@@ -32,17 +32,19 @@ if [[ -z "${_ROCM_DIR}" ]]; then
...
+_BUILD_DIR=${BUILD_DIR:-$PWD}
+export _INSTALL_DIR=${INSTALL_DIR:-$_BUILD_DIR/install}
+echo "rocSHMEM dependencies UCX and Open MPI will install in $_INSTALL_DIR"
 
+export _DEPS_SRC_DIR=$_BUILD_DIR/deps-src
 mkdir -p $_DEPS_SRC_DIR
 
 #Adjust branches and installation location as necessary
-export _UCX_INSTALL_DIR=$_INSTALL_DIR/ucx
+export _UCX_INSTALL_DIR=${INSTALL_DIR:-$_INSTALL_DIR/ucx}
```
omor1, line 43: Maybe have a separate `UCX_INSTALL_DIR` and `OMPI_INSTALL_DIR` env vars that can be set, defaulting to `${_INSTALL_DIR}/ucx` (and `${_INSTALL_DIR}/ompi` if unset? Since `_INSTALL_DIR` is set to `${INSTALL_DIR}` if that exists anyway. Users do not necessarily want those installed in the same directory when `INSTALL_DIR` is set.
```suggestion
export _UCX_INSTALL_DIR=${UCX_INSTALL_DIR:-${_INSTALL_DIR}/ucx}
```

abouteiller: I did that change because some users forgot to set LD_LIBARY_PATH and would get a different ucx than what ompi is built with. Installing them in the same directory helps removing that user-error since mpiexec prefix_by_default will include the proper LD_LIBRARY_PATH for both libmpi and libucx.

Yiltan (approved):

PR #303, opened 2025-10-30
Alltoall linear parallel Optimization

Yiltan:
Parallelize the alltoall operation if we have multiple waves to use

Depends on #299 for performance

### DIFF
--- a/src/gda/context_gda_device_coll.cpp
+++ b/src/gda/context_gda_device_coll.cpp
@@ -37,6 +37,9 @@ __device__ void GDAContext::internal_direct_barrier(int pe, int PE_start,
   if (pe == PE_start) {
     // Go through all PE offsets (except current offset = 0)
     // and wait until they all reach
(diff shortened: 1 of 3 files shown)


src/gda/context_gda_device_coll.cpp
```diff
@@ -70,6 +70,57 @@ __device__ void GDAContext::internal_direct_barrier(int pe, int PE_start,
   }
 }
 
+__device__ void GDAContext::internal_direct_barrier_wg(int pe, int PE_start,
```
abouteiller, line 73: this change is part of the other PR?

abouteiller (approved):

src/gda/context_gda_device_coll.cpp
Yiltan: Yup its one commit in addition to the other PR, its not based against develop

I could probably clean it up as it is independent, I was just going to wait for PR299 first then this

PR #304, opened 2025-10-31
Remove LOAD/STORE macros, explode store_asm into store_asmN where N=2..8

abouteiller:
## Motivation

1. Remove unused code
2. Enable removing switch from critical path in memcpy_lane/wave/wg?

## Technical Details

Remove LOAD-STORE macros and replace usage of STORE macro with its definition in HDP policy
Split `store_asm` into its component functions `store_asm{2,4,8}` so that we can call the right one from a flattened memcpy_lane (separate pr)

## Test Plan

Run IPC testers

### DIFF
--- a/src/assembly.hpp
+++ b/src/assembly.hpp
@@ -182,49 +182,61 @@ __device__ __forceinline__ void __roc_flush() {
 #endif
 }
 
+__device__ __forceinline__ void store_asm2(uint16_t& val, uint16_t* dst) {
+#if defined(__gfx906__)
+#endif
+#if defined(__gfx908__)
+#endif
+#if defined(__gfx90a__) || defined (__gfx1100__)
+  asm volatile("flat_store_short %0 %1 glc slc" : : "v"(dst), "v"(val));
+#endif
+#if defined(__gfx942__) || defined(__gfx950__)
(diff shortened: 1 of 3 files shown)


src/assembly.hpp
```diff
@@ -182,49 +182,61 @@ __device__ __forceinline__ void __roc_flush() {
...
 #endif
+}
+
+__device__ __forceinline__ void store_asm(uint8_t* val, uint8_t* dst,
+                                          int size) {
+  switch (size) {
+    case 2: {
+      uint16_t val16{*(reinterpret_cast<uint16_t*>(val))};
+      store_asm2(val16, (uint16_t*)dst);
+      break;
```
Yiltan, line 230: my cpp knowledge isn't quiet there but would this be beneficial (or something similar)?

```

template<typename T>
__device__ __forceinline__ void store_asm(T* val, T* dst, int size) {
   store_asm<size>(val, (T*)dst);
 }

```

@omor1 you'll probably know the answer to this

dsidler: you can do `sizeof(T)`, no need to pass `size`

ammallya:

PR #318, opened 2025-11-12
Feat/multi qps per endpoint

avinashkethineedi:
## Motivation
- This PR adds support for configurable QP allocation per PE in the GDA context.
- It introduces environment variables to control the number of QPs for the default and user contexts.
- Previously, the number of QPs per PE was fixed to 1.

AIROCSHMEM-7

#### Environment Variables
- Added two new environment variables to configure QP allocation:
  - `ROCSHMEM_GDA_NUM_QPS_PER_PE_DEFAULT_CTX`: Number of QPs per PE in the default context.
  - `ROCSHMEM_GDA_NUM_QPS_PER_PE_USR_CTX`: Number of QPs per PE in each user context.
- Introduced `get_qp_index()` device method to compute QP index using atomic counter
  for round-robin access.
## Test Plan
- Verify correctness by initializing multiple contexts with varying environment variable values.
- Verifiy functionality and correctness across all three NIC types: mlx, bnxt, and ionic.
- Measure and validate performance improvements with configurable QP allocation under different workloads.

## Submission Checklist

- [x] Code compiles successfully
- [x] All relevant unit tests pass
- [ ] Verified behavior under multiple context configurations
- [x] Verified functionality and performance improvements with mlx
- [ ] Verified functionality and performance improvements with bnxt
- [ ] Verified functionality and performance improvements with ionic

PR #319, opened 2025-11-12
[GDA] Alltoall optimization - single warp

Yiltan:
## Motivation

- Allow each thread to write to its own QP
- Allow for the execution of alltoall/barrier in a single warp

Conflicts with #318, the two PRs need to be merged in some order

### DIFF
--- a/src/gda/context_gda_tmpl_device.hpp
+++ b/src/gda/context_gda_tmpl_device.hpp
@@ -648,19 +648,21 @@ __device__ void GDAContext::alltoall_linear_thread_puts(rocshmem_team_t team, T 
   long *pSync = team_obj->alltoall_pSync;
   int my_pe_in_team = team_obj->my_pe;
 
+  int tid = get_flat_block_id();
   int wf_id = get_flat_block_id() / WF_SIZE;
   int wf_count = (int) ceil((double)get_flat_block_size() / (double)WF_SIZE);
   bool wf_leader = 0 == get_active_lane_num();
 
   // Have each PE put their designated data to the other PEs
-  for (int j = wf_id; j < pe_size; j+= wf_count) {
+  for (int j = tid; j < pe_size; j+= WF_SIZE) {
     int dest_pe = team_obj->get_pe_in_world(j);
-    put_nbi_wave(&dst[my_pe_in_team * nelems], &src[j * nelems], nelems, dest_pe);
+    uint64_t L_offset = reinterpret_cast<char*>(&dst[my_pe_in_team * nelems]) - base_heap[my_pe];
+    qps[dest_pe].put_nbi_single(base_heap[dest_pe] + L_offset, &src[j * nelems], nelems * sizeof(T), dest_pe);
   }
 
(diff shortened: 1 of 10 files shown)


src/gda/context_gda_tmpl_device.hpp
```diff
@@ -604,7 +605,11 @@ __device__ void GDAContext::internal_broadcast(T *dst, const T *src, int nelems,
 template <typename T>
 __device__ void GDAContext::alltoall(rocshmem_team_t team, T *dst,
                                      const T *src, int nelems) {
-  alltoall_linear_thread_puts(team, dst, src, nelems);
+  if (gda_provider_ == GDAProvider::BNXT) {
+    alltoall_linear_thread_puts(team, dst, src, nelems);
```
nusislam, line 609: Why is this not applicable for other NICs?

Yiltan: - Should work on IONIC (untested, so I've left it out)
- There is a pretty major PR in the works for CX7, adding the logic to support multiple threads writing into their own QP would create a major conflict for them. It would be best to add this support once the other PR is merged

Yiltan:
Fixed the bug where alltoall would hang if the thread count was fewer than the number of GPUs. Ready to review

src/gda/context_gda_tmpl_device.hpp
```diff
@@ -637,6 +641,36 @@ __device__ void GDAContext::alltoall_linear(rocshmem_team_t team, T *dst,
...
+  int my_pe_in_team = team_obj->my_pe;
+
+  int tid = get_flat_block_id();
+  int step_size = min(get_flat_block_size(), WF_SIZE);
+
+  // Have each PE put their designated data to the other PEs
+  for (int j = tid; j < pe_size; j+= step_size) {
+    int dest_pe = team_obj->get_pe_in_world(j);
+    uint64_t L_offset = reinterpret_cast<char*>(&dst[my_pe_in_team * nelems]) - base_heap[my_pe];
+    qps[dest_pe].put_nbi_single(base_heap[dest_pe] + L_offset, &src[j * nelems], nelems * sizeof(T), dest_pe);
```
edgargabriel, line 662: Do I understand it correctly that with this approach taking advantage of IPC is not an option?

edgargabriel (approved):

src/gda/context_gda_tmpl_device.hpp
Yiltan: Yes, it does not take advantage of IPC.

When calling into the put function `isIpcAvailable` causes wave divergence. I.e thread 0-7 will be IPC and thread 8-63 will be GDA. 

I found the performance to be better only using GDA. Using the regular putmem would add 20us or so

There is a longer term solution we could take to take advantage of IPC that would require us to use two waves. Something like:

```
if (wave_id == 0) { // Inter-Node
  gda_put(...);
}
else { // Intra-Node
  ipc_put(...);
}
```

PR #331, opened 2025-11-20
Refactor: modularize RMA and AMO WQE posting functions

avinashkethineedi:
## Motivation

- Modularize and simplify the WQE posting code in MLX code path

## Technical Details

- Extracted shared logic for SQ/CQ waiting, doorbell ringing, and WQE construction.
- Updated both RMA and AMO paths to use the new helpers.
- No functional or behavioral changes.

## Test Plan

- Build and run existing tests.
- Verify RMA and AMO operations with functional tests.
- Compare behavior with baseline to ensure no differences.

## Test Result

- All tests passed.
- RMA and AMO behavior matches baseline.
- No performance regressions observed.

## Submission Checklist

- [x] Code compiles
- [x] Tests pass
- [x] No functional changes
- [x] Refactor only

### DIFF
--- a/src/gda/mlx5/queue_pair_mlx5.cpp
+++ b/src/gda/mlx5/queue_pair_mlx5.cpp
@@ -122,41 +122,47 @@ __device__ void QueuePair::mlx5_quiet() {
   }
 }
 
-__device__ void QueuePair::mlx5_post_wqe_rma(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode) {
-  uint64_t activemask = get_active_lane_mask();
-  uint8_t num_active_lanes = get_active_lane_count(activemask);
-  uint8_t my_logical_lane_id = get_active_lane_num(activemask);
-  bool is_leader{my_logical_lane_id == 0};
-  const uint64_t leader_phys_lane_id = get_first_active_lane_id(activemask);
-  uint8_t num_wqes{num_active_lanes};
-  uint64_t wave_sq_counter{0};
-
-  if (is_leader) {
-    wave_sq_counter = __hip_atomic_fetch_add(&sq_posted, num_wqes, __ATOMIC_SEQ_CST, __HIP_MEMORY_SCOPE_AGENT);
-  }
-  wave_sq_counter = __shfl(wave_sq_counter, leader_phys_lane_id);
-  uint64_t my_sq_counter = wave_sq_counter + my_logical_lane_id;
-  uint64_t my_sq_index = my_sq_counter % sq_wqe_cnt;
-
+__device__ __forceinline__ void QueuePair::mlx5_wait_for_free_sq_slots(
+    uint64_t wave_sq_counter, uint8_t num_active_lanes) {
   while (true) {
-    uint64_t db_touched = __hip_atomic_load(&sq_db_touched, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
-    uint64_t sunk = __hip_atomic_load(&sq_sunk, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
-    int64_t num_active_sq_entries = db_touched - sunk;
+    uint64_t db_touched = __hip_atomic_load(&sq_db_touched,
+                          __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+
+    uint64_t sunk       = __hip_atomic_load(&sq_sunk,
+                          __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT);
+
+    int64_t num_active_sq_entries =
+        static_cast<int64_t>(db_touched) -
+        static_cast<int64_t>(sunk);
+
     if (num_active_sq_entries < 0) {
       continue;
     }
-    uint64_t num_free_entries = min(sq_wqe_cnt, cq_cnt) - num_active_sq_entries;
-    uint64_t num_entries_until_wave_last_entry = wave_sq_counter + num_active_lanes - db_touched;
+
+    uint64_t num_free_entries =
+        min(sq_wqe_cnt, cq_cnt) -
+        static_cast<uint64_t>(num_active_sq_entries);
+
+    uint64_t num_entries_until_wave_last_entry =
+        wave_sq_counter + num_active_lanes - db_touched;
+
     if (num_free_entries > num_entries_until_wave_last_entry) {
       break;
     }
+
     mlx5_quiet();
   }
+}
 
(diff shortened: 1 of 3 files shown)


src/gda/queue_pair.hpp
```diff
@@ -164,15 +164,59 @@ class QueuePair {
...
+      uintptr_t *raddr, uint8_t opcode, int64_t atomic_data,
+      int64_t atomic_cmp, bool fetching, uint64_t *wave_fetch_atomic,
+      uint8_t my_logical_lane_id);
+
+  __device__ __forceinline__ uint64_t*
+  mlx5_allocate_wave_fetching_atomic_buffer(uint64_t wave_sq_counter,
+      bool is_leader, uint64_t leader_phys_lane_id);
+
+  __device__ __forceinline__ void
+  mlx5_ring_wave_doorbell(uint64_t wave_sq_counter, uint8_t num_wqes);
```
Yiltan: Having wave as an infix is different than the current patten we follow in rocSHMEM as we usually have `_wave` as a suffix. Same comment applies to other functions

I would consider following the current pattern or what about something like this:

```
<template Collectivity cy>
__device__ __forceinline__ void mlx5_ring_doorbell(...)
```

1) Uses existing "Collectivity" term used by IONIC, there could be a better term but this uses existing code

2) Having the argument `uint8_t num_wqes` will not translate to other NICs, this is an implementation choice by the mlx5 code. For ringing the doorbell, we can either pass the slot id that we want to write at or  ring doorbell should just take the value to write

src/gda/queue_pair.hpp
```diff
@@ -164,15 +164,59 @@ class QueuePair {
...
 #if defined(GDA_MLX5)
-  __device__ uint64_t mlx5_post_wqe_amo(int pe, int32_t size, uintptr_t *raddr, uint8_t opcode, int64_t atomic_data, int64_t atomic_cmp, bool fetch);
-  __device__ void mlx5_post_wqe_rma(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode);
-  __device__ void mlx5_quiet();
+  __device__ __forceinline__ void
+  mlx5_wait_for_free_sq_slots(uint64_t wave_sq_counter,
+      uint8_t num_active_lanes);
+
+  __device__ __forceinline__ void
+  mlx5_wait_for_db_touched_eq(uint64_t target_sq_counter);
```
Yiltan, line 188: `eq`? I don't think we are using event queues are we?

src/gda/queue_pair.hpp
```diff
@@ -164,15 +164,59 @@ class QueuePair {
...
+
+  __device__ __forceinline__ void
+  mlx5_build_rma_wqe(uint64_t my_sq_counter, uint64_t my_sq_index,
+      uintptr_t *laddr, uintptr_t *raddr, int32_t size, uint8_t opcode);
+
+  __device__ __forceinline__ void
+  mlx5_build_amo_wqe(uint64_t my_sq_counter, uint64_t my_sq_index,
```
Yiltan, line 195: This comment can be applied in a future PR as to not block this effort. My comments here are suggestions on how we can aid the unification of the writing of WQEs for different NIC vendors

1) Having both `my_sq_counter` and `my_sq_index` is a design choice in the mlx5 path. We only need the slot id to write to an SQ, this would make it easier to unify this operation between backends. 

2) The parameter `my_logical_lane_id` is used before this call to select the correct `sq_counter/idx` but we use it again inside inside to select the `wave_fetch atomic` (I assume this is the return address). I think we should do this calculation outside of the function.

3) `bool fetching` can probably probably be templated/constexpr as we know that at compile time 

Similar comments apply to `mlx5_build_rma_wqe`

src/gda/queue_pair.hpp
```diff
@@ -164,15 +164,59 @@ class QueuePair {
    * @param[in] raddr Remote address.
    * @param[in] opcode Operation to be performed.
    */
-  __device__ __attribute__((noinline)) void post_wqe_rma(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode, Collectivity cy);
-  __device__ __attribute__((noinline)) void post_wqe_rma_turn(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode, Collectivity cy);
-  __device__ __attribute__((noinline)) void post_wqe_rma_single(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode);
-  __device__ __attribute__((noinline)) void post_wqe_rma_mt(int pe, int32_t size, uintptr_t *laddr, uintptr_t *raddr, uint8_t opcode);
+  __device__ __attribute__((noinline)) void
+  post_wqe_rma(int pe, int32_t size, uintptr_t *laddr,
+      uintptr_t *raddr, uint8_t opcode, Collectivity cy);
```
Yiltan:
```
  __device__ __attribute__((noinline)) void
  post_wqe_rma(int pe, int32_t size, uintptr_t *laddr,
               uintptr_t *raddr, uint8_t opcode, Collectivity cy);
```

src/gda/mlx5/queue_pair_mlx5.cpp
```diff
@@ -122,147 +122,212 @@ __device__ void QueuePair::mlx5_quiet() {
...
-    seg_build.update_data_seg(wave_fetch_atomic + my_logical_lane_id, 8, fetching_atomic_lkey);
+    seg_build.update_data_seg(wave_fetch_atomic + my_logical_lane_id, 8,
+                              fetching_atomic_lkey);
   } else {
     seg_build.update_data_seg(nonfetching_atomic, 8, nonfetching_atomic_lkey);
   }
-  __atomic_signal_fence(__ATOMIC_SEQ_CST);
+}
 
+__device__ uint64_t QueuePair::mlx5_post_wqe_amo(int pe, int32_t size,
```
Yiltan: `int pe` parameter is unused, done in a few other places too. This wasn't introduced in this PR but if we are refactoring this function, its a good time to remove it.

src/gda/queue_pair.hpp
avinashkethineedi: `eq` as in `equal_to`

Yiltan (dismissed):

src/gda/queue_pair.hpp
avinashkethineedi: 1. `my_sq_counter` is still required for the table that is maintained in the mlx code path
2. Updateed the address calculation outside of the function
3. It requires changes in the code path of other NICs as well, will differ this to different PR

Yiltan (dismissed):

Yiltan (approved):

PR #332, opened 2025-11-25
IPC ldstore optimizations

akolliasAMD:
## Motivation

Testing different methods for memcpy

### DIFF
--- a/src/assembly.hpp
+++ b/src/assembly.hpp
@@ -230,7 +230,7 @@ __device__ __forceinline__ void store_asm(uint8_t* val, uint8_t* dst,
       asm volatile("flat_store_short %0 %1 glc slc" : : "v"(dst), "v"(val16));
 #endif
 #if defined(__gfx942__) || defined(__gfx950__)
-      asm volatile("flat_store_short %0 %1 sc0 sc1" : : "v"(dst), "v"(val16));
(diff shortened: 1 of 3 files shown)


src/assembly.hpp
```diff
@@ -230,15 +230,15 @@ __device__ __forceinline__ void store_asm(uint8_t* val, uint8_t* dst,
       asm volatile("flat_store_short %0 %1 glc slc" : : "v"(dst), "v"(val16));
 #endif
 #if defined(__gfx942__) || defined(__gfx950__)
-      asm volatile("flat_store_short %0 %1 sc0 sc1" : : "v"(dst), "v"(val16));
+      __builtin_nontemporal_store(val16, (int16_t *)dst);
```
omor1, line 233: Can we get rid of all the arch-specific stuff and just use `__builtin_nontemporal_load` and `__builtin_nontemporal_store` directly? Or does it not generate the right code sequences for gfx90a/gfx1201?

It would also be nice if we can turn `size` into a compile-time constant (templating, maybe?), though that might take some more cleverness (template wizardry) to programmatically unroll the loops in util.hpp and the compiler is probably able to constant-fold this anyway.

omor1 (approved):

ammallya:

PR #333, opened 2025-11-25
Add host API for alltoallmem_on_stream collective operation

amd-arozanov:
## Motivation
(AI Assist)

This PR adds a host-side asynchronous stream-based API for the alltoall collective operation. 

The new rocshmem_alltoallmem_on_stream host function enables users to enqueue alltoall operations on HIP streams, allowing parallel execution across multiple streams and better overlap of communication and computation.

Ticket: https://amd-hub.atlassian.net/browse/AIROCSHMEM-202

## Technical Details

**Function signature:**
`__host__ void rocshmem_alltoallmem_on_stream(rocshmem_team_t team, void *dest, const void *source, size_t size, hipStream_t stream)`

**Implementation Details:**
- Support added across all backends (GDA, IPC, RO)
- The implementation creates a separate context for each workgroup to avoid contention on the default context, allowing parallel execution across multiple streams
- Uses dynamic block size determination via HIP occupancy API to optimize kernel launch configuration
- If `stream` is `nullptr`, the operation defaults to `hipStreamDefault`

**Bug Fix:**
- Fixed incorrect sync buffer size allocation for alltoall in GDA and IPC backends (commit 069c6d4)
  - Changed from `ROCSHMEM_BCAST_SYNC_SIZE` to `ROCSHMEM_ALLTOALL_SYNC_SIZE` 

**Testing:**
- Added new functional test `team_alltoallmem_on_stream_tester`

**Performance Considerations:**
- Compared to the device-side `rocshmem_alltoall_wg()` API, the host-side stream API has additional overhead:
  - **Kernel launch overhead**: Each call requires launching a kernel (`rocshmem_alltoallmem_kernel`), which adds latency for kernel launch and setup
  - **Context creation/destruction**: The kernel creates and destroys a workgroup context for each invocation (`rocshmem_wg_team_create_ctx`/`rocshmem_wg_ctx_destroy`). Note: This overhead could potentially be reduced by caching contexts in the team structure when using one team per workgroup, but this optimization is not currently implemented.

## Test Plan

rocSHMEM functional testing

## Test Result

All tests passed except: 
- teamreduction_n2_w1_z1_32768B. Seems it fails in other PRs too
- ubuntu22 && 8gfx950-no-interconnect configuration is not running

## Submission Checklist

### DIFF
--- a/docs/api/coll.rst
+++ b/docs/api/coll.rst
@@ -88,6 +88,29 @@ This function must be called as a work-group collective.
 
 Valid ``TYPENAME`` and ``TYPE`` values are listed in :ref:`RMA_TYPES`.
 
+ROCSHMEM_ALLTOALLMEM_ON_STREAM
+-------------------------------
+
+.. cpp:function:: __host__ void rocshmem_alltoallmem_on_stream(rocshmem_team_t team, void *dest, const void *source, size_t size, hipStream_t stream)
+
+  :param team:   The team participating in the collective.
+  :param dest:   Destination address. Must be an address on the symmetric heap.
+  :param source: Source address. Must be an address on the symmetric heap.
+  :param size:   Number of bytes to transfer per pair of PEs.
+  :param stream: HIP stream on which to enqueue the operation.
+  :returns:      None.
+
+**Description:**
+This routine enqueues an alltoall collective operation on a HIP stream. The function
+exchanges a fixed amount of contiguous data blocks between all pairs of PEs participating
+in the collective routine. The operation is enqueued on the specified stream and will
+execute asynchronously. The caller must synchronize the stream (e.g., using
+``hipStreamSynchronize``) to ensure completion.
+
+This function creates a separate context for each workgroup to avoid contention on the
+default context, allowing parallel execution across multiple streams.
+If ``stream`` is ``nullptr``, the operation will use ``hipStreamDefault``.
+
 ROCSHMEM_BROADCAST
 ------------------
 
--- a/include/rocshmem/rocshmem.hpp
+++ b/include/rocshmem/rocshmem.hpp
@@ -350,6 +350,22 @@ __host__ void rocshmem_barrier_all();
 __host__ void rocshmem_barrier_all_on_stream(hipStream_t stream);
 
 /**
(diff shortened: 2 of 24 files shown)


amd-arozanov:
Could someone review this PR? I've prepared the next one with more new APIs: https://github.com/ROCm/rocSHMEM/pull/340

docs/api/coll.rst
```diff
@@ -88,6 +88,29 @@ This function must be called as a work-group collective.
...
+  :returns:      None.
+
+**Description:**
+This routine enqueues an alltoall collective operation on a HIP stream. The function
+exchanges a fixed amount of contiguous data blocks between all pairs of PEs participating
+in the collective routine. The operation is enqueued on the specified stream and will
+execute asynchronously. The caller must synchronize the stream (e.g., using
+``hipStreamSynchronize``) to ensure completion.
+
+This function creates a separate context for each workgroup to avoid contention on the
```
Yiltan, line 110: I'm not sure about the wording here. "creates a separate ctx for each workgroup". However in line 361 in host.cpp we only launch with one workgroup (`gridSize(1)`). 

We only launch with a single workgroup, so I'm not sure how much contention this would cause on the default context. Its unclear if posting on a user context would cause any ordering issues. I think this needs another pair of eyes on from the rocSHMEM team.

Otherwise the PR does look good to me

amd-arozanov: Here I meant that if you run rocshmem_alltoallmem_on_stream several times, then it creates new context for each invocation. 

Like I do it in the test:
https://github.com/ROCm/rocSHMEM/blob/f12fe0dcbc3e99a1778d6407d43a9ad7f4b5b807/tests/functional_tests/team_alltoallmem_on_stream_tester.cpp#L174

Here I call rocshmem_alltoallmem_on_stream several times in different streams and with different teams. And I see that each call executes in parallel with others.

Yiltan (approved):

edgargabriel (approved):

edgargabriel:
Testing looks good, the failure on gfx942 is not related to this PR, everything else passes.

amd-arozanov:
Can we merge this PR?

PR #335, opened 2025-11-26
Add rocshmem_int64_p

Yiltan:
The API was declared internally but we did not expose it

### DIFF
--- a/docs/api/rma.rst
+++ b/docs/api/rma.rst
@@ -218,7 +218,7 @@ The following table lists the supported RMA data types:
       - No
     * - int64_t
       - int64
-      - No
+      - Yes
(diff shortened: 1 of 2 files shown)


docs/api/rma.rst
abouteiller, line 216: PR is simple enough, great, is there a reason to not extend this to all intX_t types in this table?

abouteiller (approved):

docs/api/rma.rst
Yiltan: Only focused on the API for the request, but there is no technical reasons why we cannot implement the other datatypes. I'm not 100% sure that they are all defined internally but it would be simple if they are not

PR #336, opened 2025-11-27
added a way to print functional test results into an easy to deal wit…

akolliasAMD:
simple changes to create one more usable log

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -164,12 +164,29 @@ ExecTest() {
     TEST_LOG_NAME+=_"$MAX_MSG_SIZE"B
   fi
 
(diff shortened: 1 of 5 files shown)


scripts/functional_tests/driver.sh
```diff
@@ -660,7 +677,8 @@ ValidateLogDir() {
 APP=$1
 TEST=$2
 LOG_DIR=$3
-HOSTFILE=$4
+CSVPRINT=$4
+HOSTFILE=$5
```
Yiltan, line 681: 1) We should probably update the help message here: https://github.com/ROCm/rocSHMEM/blob/develop/scripts/functional_tests/driver.sh#L639

2) With this modification, is it possible to run multi-node without the CSV output? (we may have to use optarg rather than our current "hack")

PR #337, opened 2025-11-27
Update docs for GDA

Yiltan:
Update docs for GDA

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -15,7 +15,7 @@ Requirements
 
 * ROCm 6.4.0 or later, including the :doc:`HIP runtime <hip:index>`. For more information, see `ROCm installation for Linux <https://rocm.docs.amd.com/projects/install-on-linux/en/latest/>`_.
 
-* The following AMD GPUs have been fully tested for compatibility with rocSHMEM: 
+* The following AMD GPUs have been fully tested for compatibility with rocSHMEM:
 
   * MI250X
 
@@ -29,7 +29,7 @@ Requirements
 
 * ROCm-aware Open MPI and UCX. For more information, see :ref:`install-dependencies`.
 
-* Inter-node communication requires MPI, and is tested with Open MPI and CX7 Infiniband NICs.
+* Inter-node communication requires AMD Pollara IONIC, Broadcom Thor 2, or CX7 Infiniband NICs.
 
 Available network backends
 --------------------------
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -29,22 +29,23 @@ Requirements
 
 * ROCm-aware Open MPI and UCX. For more information, see :ref:`install-dependencies`.
 
-* Inter-node communication requires MPI, and is tested with Open MPI and CX7 Infiniband NICs.
+* Inter-node communication requires AMD Pollara IONIC, Broadcom Thor 2, or CX7 Infiniband NICs.
 
 Available network backends
 --------------------------
 
 rocSHMEM supports two network backends:
```
gaoikawa: 'two' -> we should change this to 'the following' (or something that doesn't require us to keep updating the number of backends.

docs/install.rst
```diff
@@ -29,22 +29,23 @@ Requirements
...
 
 * The **IPC (Inter-Process Communication)** backend enables fast communication between GPUs on the same host using ROCm inter-process mechanisms. It does not support inter-node communication.
-* The **RO (Reverse Offload)** backend enables communication between GPUs on different nodes through a NIC, using a host-based proxy to forward communication orders to and from the GPU. In this release, RO is the only inter-node communication backend and is built on an MPI-RMA compatibility layer.
+* The **RO (Reverse Offload)** backend enables communication between GPUs on different nodes through a NIC, using a host-based proxy to forward communication orders to and from the GPU. RO is built on an MPI-RMA compatibility layer.
+* The **GDA (GPU Direct Async)** backend enables communication between GPUs on different nodes through a NIC. In this backend the GPU directly interacts with the NIC with no host involvement.
```
gaoikawa: > with no host involvement

Can you add 'CPU' to it - just to be crystal clear.

'with no host CPU involvement'.

Thanks.

docs/install.rst
```diff
@@ -64,7 +65,18 @@ On Ubuntu, you can install rocSHMEM by running:
...
 
-rocSHMEM requires ROCm-Aware Open MPI and UCX. Other MPI implementations, such as MPICH, have not been fully tested.
+GDA NIC Dependencies
+^^^^^^^^^^^^^^^^
+
+- GDA on Mellanox NICs should work on any recent version of rdma-core.
+- GDA on Broadcom Thor requires driver version 233.2.108.0 and firmware version 233.2.104.0 or later.
+
+
+Building rocSHMEM with MPI
```
gaoikawa: I'm not sure how best to phrase this, or what section it should go under, but can we mention explicit instructions on how to build rocSHMEM without MPI?   I forget the new build args / steps.

And if we mention that, can we also mention that some functional tests won't work without MPI.

docs/install.rst
```diff
@@ -29,22 +29,23 @@ Requirements
 
 * ROCm-aware Open MPI and UCX. For more information, see :ref:`install-dependencies`.
```
gaoikawa: > ROCm-aware Open MPI and UCX.

Can we update this to say Open MPI is optional now (is that only for GDA?)  (I forget if that includes not needing UCX also for GDA...)

gaoikawa (changes requested):
Requested a few changes - thanks.

gaoikawa (approved):
Changes look good - thank you @Yiltan !

PR #343, opened 2025-12-02
Functional tests without MPI support

abouteiller:
## Motivation

* Enable building and running functional tests without MPI 
* build without an external MPI
* run without loading libmpi or calling MPI_INIT
* run RO with ROCSHMEM_TEST_UUID=1 initialization
* run TEST_UUID with GDA device-interface and RO host-interface

## Technical Details

* PMIx must be finalized before calling MPI_Init (presumably a defect in Open MPI, should be possible to have two PMIx clients registered at the same time).
* RO initialization from bootstrap initialization testing found bugs and issues
* Adding RO backend_can_run to check if loading MPI is possible
* bugfix: if libibverbs.so is not found we should not crash, just GDA backend_can_run return false.
* bugfix: we cannot unload mlx5_dv (and similar) before we run ibv_device_list as that leaves libibverbs in a broken state. 

## Test Plan

Tested on Banff-RAD in the above configurations. 

* RO w/o EXTERNAL_MPI with rocshmem_init works
* GDA w/o EXTERNAL_MPI with rocshmem_init works
* RO w/o EXTERNAL_MPI with rocshmem_init_attr works
* GDA w/o EXTERNAL_MPI with rocshmem_init_attr works

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -164,7 +164,6 @@ if (NOT BUILD_TESTS_ONLY)
     set(HAVE_EXTERNAL_MPI ON)
   else()
     set(HAVE_EXTERNAL_MPI OFF)
-    set(BUILD_FUNCTIONAL_TESTS OFF)
     set(BUILD_UNIT_TESTS OFF)
   endif()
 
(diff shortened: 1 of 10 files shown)


edgargabriel (approved):

abouteiller:
CI errors appear unrelated plan is to merge on friday evening.

PR #350, opened 2025-12-08
Add default context alltoall API

Yiltan:
Add default context alltoall

### DIFF
--- a/src/rocshmem_gpu.cpp
+++ b/src/rocshmem_gpu.cpp
@@ -675,8 +675,8 @@ __global__ ATTR_NO_INLINE void rocshmem_alltoallmem_kernel(rocshmem_team_t team,
 
   // Call device alltoall function with created context and provided team
   // Using char type since size is in bytes (1 byte per element)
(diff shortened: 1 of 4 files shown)


thomas-huber (dismissed):

thomas-huber:
compile fix: 

Change line 673 in `rocshmem_gpu.cpp` to `rocshmem_ctx_alltoall_wg`

src/rocshmem_gpu.cpp
```diff
@@ -529,15 +529,24 @@ __device__ void rocshmem_broadcast_wg(rocshmem_ctx_t ctx,
 }
 
 template <typename T>
-__device__ void rocshmem_alltoall_wg(rocshmem_ctx_t ctx,
```
abouteiller, line 537: This is a backward incompatible change, will require bumping version major.

abouteiller: My bad this is not externally visible in API

abouteiller (approved):

avinashkethineedi (approved):

PR #352, opened 2025-12-09
Reenable building test-only with external MPI

abouteiller:
## Motivation

When building test-only with rocshmem configuration 'external MPI' the build stage of the testers would not find dependent packages as declared by the INTERFACE of rocshmem.

## Technical Details

rocshmem-config.cmake would add MPI::MPI_CXX and other dependencies to the library interface when built with external MPI. The dependee must do a find_package for these dependencies.

## Test Plan

* [x] Can compile with rocshmem-dev (with EXTERNAL_MPI set)
* [x] Can compile with devel (no EXTERNAL MPI set)

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -29,6 +29,7 @@ cmake_minimum_required(VERSION 3.16.3 FATAL_ERROR)
 include(${CMAKE_SOURCE_DIR}/cmake/setup_project.cmake)
 project(rocshmem_examples VERSION 1.0.0 LANGUAGES CXX)
 
+find_package(MPI)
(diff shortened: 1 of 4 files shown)


examples/CMakeLists.txt
```diff
@@ -45,7 +46,7 @@ set(EXAMPLE_SOURCES
   rocshmem_put_signal_test.cc
 )
 
-if (HAVE_EXTERNAL_MPI)
+if (MPI_CXX_FOUND)
```
edgargabriel, line 49: I take it back, misread the commit,

edgargabriel (approved):

PR #327, opened 2025-11-19
AIROCSHMEM-201 - gda: add check for active interfaces when selecting the GDA backend

VadimKutovoi:
## Motivation

Ticket: AIROCSHMEM-201

## Technical Details

Add checks to ensure GDA backend only activates when:
- Direct Verbs library can be opened
- Active RDMA interfaces exist (IBV_PORT_ACTIVE)
- Device vendor matches provider (BNXT/IONIC/MLX5)

This prevents crashes when libraries are installed but hardware is
unavailable or mismatched.

## Test Plan

Default functional and manual testing.

## Test Result

Manual testing:

Case 1 (Matching provider and hardware. Broadcom + bnxt provider):

 ROCSHMEM_GDA_PROVIDER=bnxt mpirun --oversubscribe -np 2 examples/rocshmem_alltoall_test
```
Found active InfiniBand port 1 on device bnxt_re0 (vendor_id=0x14e4, state=4, phys_state=5)
GDABackend::backend_can_run returned success
Initializing GDA backend using MPI
Found active InfiniBand port 1 on device bnxt_re0 (vendor_id=0x14e4, state=4, phys_state=5)
GDABackend::backend_can_run returned success
Initializing GDA backend using MPI
GPU Device id: 0 closest NIC id : 0 name: bnxt_re0
GPU Device id: 1 closest NIC id : 2 name: bnxt_re2
```

Case 2 (Provider and hardware mismatch. Broadcom + mlx5 provider):

 ROCSHMEM_GDA_PROVIDER=mlx5 mpirun --oversubscribe -np 2 examples/rocshmem_alltoall_test
```
Found environment variable ROCSHMEM_GDA_PROVIDER, value is mlx5
Found environment variable ROCSHMEM_GDA_PROVIDER, value is mlx5
Skipping device bnxt_re0 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re1 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re2 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re3 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re4 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re0 with vendor_id=0x14e4 (not MLX5/Mellanox)
Skipping device bnxt_re5 with vendor_id=0x14e4 (not MLX5/Mellanox)
[...]

PR #357, opened 2025-12-09
remove cq lock

Yiltan:
## Motivation

PR #327 indicated that when using collapsed CQE mode, we do not need locks around the CQ

## Technical Details

Removed locks from the bnxt code path

## Test Plan

- [x] rocSHMEM functional tests
- [x] DeepEP validation

Some minor performance improvement was noted in both tests

### DIFF
--- a/src/gda/bnxt/queue_pair_bnxt.cpp
+++ b/src/gda/bnxt/queue_pair_bnxt.cpp
@@ -177,8 +177,6 @@ __device__ void QueuePair::poll_cq_until(uint32_t requested_available_slots) {
 
   sq_depth = sq.depth;
 
-  aquire_lock(&cq.lock);
-
   do {
     cqe = (struct bnxt_re_req_cqe *) cq.buf;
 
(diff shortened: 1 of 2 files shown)


omor1:
The only modification that would potentially need to be protected by locks are updates to `sq.head`; however, I don't think that this is read anywhere, so even though the writes are racy it's fine. LGTM, though I'd recommend removing `sq.head` if we don't use it.

I'll note that there are some potential ordering issues inside `quiet()` between updates to `cqe->con_indx` (by the HCA) and to `sq.tail` (by other threads inside `post_wqe_*()`). I haven't fully analyzed whether it's possible, but the concern would be a scenario where we observe an update to `cqe->con_indx` prior to the corresponding update to `sq.tail`, i.e. the consumer (the HCA) appears to overtake the producer(s) (GPU threads in `post_wqe_*()`). However, I think even if this occurs, it will simply appear that the queue is very full and so in all likelihood the loop will simply continue for another iteration.

src/gda/bnxt/queue_pair_bnxt.cpp
omor1, line 191: Removing the lock makes these writes racy, but I don't think we use `sq.head` anywhere else i.e. it isn't read. I'd recommend removing the racy write to `sq.head`, and possibly the variable itself if it's not needed.

omor1 (approved):

PR #360, opened 2025-12-10
use 64 threads for reduction test

edgargabriel:
## Motivation

use 64 threads for reduction test, much faster with IPC backend.

## Submission Checklist

### DIFF
--- a/scripts/functional_tests/driver.sh
+++ b/scripts/functional_tests/driver.sh
@@ -450,7 +450,7 @@ TestColl() {
   ExecTest  "fcollect"         2       1            1         512
   ExecTest  "fcollect"         2       1            1         32768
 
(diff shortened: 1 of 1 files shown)


Yiltan (dismissed):

abouteiller (dismissed):

abouteiller:
should we do that for all collective that have actual data exchange?

edgargabriel:
> should we do that for all collective that have actual data exchange?

probably a good idea. I was just thinking about the reduction test because that was the one I was looking into, but would probably make sense for fcollect, bcast, alltoall as well

avinashkethineedi (dismissed):

Yiltan (approved):

abouteiller (approved):

PR #361, opened 2025-12-10
Tests/deepep pattern

abouteiller:
## Motivation

Import deepep-like (internode) testing pattern

## Technical Details

Extracted from prior works during GDA rampup and deepep evaluation

Todo: 
* [ ] cmake integration
* [ ] integration with main tester.cpp class

## Test Plan

* [ ] Add to functional tests and pass


ammallya:
Imported to ROCm/rocm-systems pr 2733

PR #364, opened 2025-12-12
[GDA/mlx5] CQ collapsing

omor1:
## Motivation

CQ collapsing has shown significant benefits for bnxt and ionic, we need to examine its use in mlx5.

AIROCSHMEM-5

## Technical Details

* The prior design inherently assumes that all CQEs will be observed: the number of observed CQEs is compared to the number of posted WQEs to determine SQ capacity. For performance, multiple threads per wavefront are used to poll the CQ.
* CQ collapsing will practically guarantee that some CQEs will be missed—you only observe the most recent CQE, with prior CQEs being overwritten. Only a single thread should be used to poll the CQ, since there's only a single CQE that needs to be polled. The prior design is thus not compatible with CQ collapsing.
* This PR replaces the QueuePair implementation with one based off the bnxt code. The code is massively simplified and now has comments explaining its behavior.
* In the future we will determine how to more closely merge the bnxt and mlx5 implementations to reduce the amount of duplicate code. This will also entail determining whether and how we can remove the current SQ lock and replace it by a lockless implementation, which is one feature that gets lost by utilizing the new design.
* Depends on PRs:
  * #368
  * #369
  * #373

## Test Plan

- [ ] Functional tests
- [ ] Microbenchmarks
- [ ] DeepEP

## Test Result

N/A, not yet completed

### DIFF
--- a/src/gda/backend_gda.cpp
+++ b/src/gda/backend_gda.cpp
@@ -1282,6 +1282,19 @@ void GDABackend::create_cqs(int cqe) {
   cq_attr.comp_mask     = IBV_CQ_INIT_ATTR_MASK_PD;
   cq_attr.parent_domain = pd_parent;
 
+  /* enable mlx5 CQ collapsing by setting CQ length to 1 and enabling CQ overrun ignore:
+   *  - mlx5 driver sets mlx5_ifc_cqc_bits::oi bit when IBV_CREATE_CQ_ATTR_IGNORE_OVERRUN is set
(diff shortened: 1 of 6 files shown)


omor1:
The branch needs to be rebased. There are also a couple minor compilation issues due to incorrect casting in the current code that I haven't gotten around to fixing yet.

I also need to double-check an assumption I had regarding how `post_wqe_amo` is called that I now think may be incorrect, but the fix should be fairly simple.

ammallya:
Imported to ROCm/rocm-systems pr 2732

PR #375, opened 2025-12-23
ipc: use fence(pe) instead of fence()

edgargabriel:
## Motivation

if possible, use fence(pe) instead of fence() in the IPC backend, and only one thread per PE needs to execute that.

## Submission Checklist


ammallya:
Imported to ROCm/rocm-systems pr 2731

PR #385, opened 2026-01-10
fix allreduce tester

edgargabriel:
## Motivation
- use the reduce_psync buffers for synchronization in allreduce, not the barrier_psync.
- execute a wwg barrier after the allreduce operation. After internal discussion it was determined that it is required for correctness.

## Submission Checklist

### DIFF
--- a/src/ipc/context_ipc_tmpl_device.hpp
+++ b/src/ipc/context_ipc_tmpl_device.hpp
@@ -174,7 +174,7 @@ __device__ void IPCContext::internal_direct_allreduce(
   int stride = team_obj->tinfo_wrt_world->stride;
   int PE_start = team_obj->tinfo_wrt_world->pe_start;
   int PE_size = team_obj->tinfo_wrt_world->size;
-  long *pSync = team_obj->barrier_pSync;
+  long *pSync = team_obj->reduce_pSync;
   T *pWrk = reinterpret_cast<T *>(team_obj->pWrk);
 
   int finish = PE_start + stride * PE_size;
@@ -195,7 +195,7 @@ __device__ void IPCContext::internal_direct_allreduce(
                     nelems * sizeof(T), i);
 
       if (is_thread_zero_in_block()) {
-        fence();
+        fence(i);
         internal_putmem(&pSync[pe], &flag_val, sizeof(*pSync), i);
       }
     }
@@ -222,7 +222,6 @@ __device__ void IPCContext::internal_direct_allreduce(
   for (int i = wg_id; i < num_pes; i += wg_size) {
     pSync[i] = ROCSHMEM_SYNC_VALUE;
   }
-  threadfence_system();
   __syncthreads();
 }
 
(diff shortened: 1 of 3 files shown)


abouteiller (approved):

src/ipc/context_ipc_tmpl_device.hpp
```diff
@@ -320,8 +319,8 @@ __device__ void IPCContext::internal_ring_allreduce(
                     chunk_size * sizeof(T), send_pe);
 
       if (is_thread_zero_in_block()) {
-        fence();
         wait_val = seg + 100;
+        fence(send_pe);
```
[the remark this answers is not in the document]
edgargabriel: The order between assigning the value to `wait_val` and the `fence(pe)` shouldn't matter, the change was more incidental.

src/ipc/context_ipc_tmpl_device.hpp
```diff
@@ -409,6 +407,7 @@ __device__ int IPCContext::reduce(rocshmem_team_t team, T *dest,
       return ROCSHMEM_ERROR;
     }
   }
+  barrier_wg(team);
```
omor1, line 410: Have you tested whether there's any major difference between having the barrier at the end versus at the start? Either should be correct.

This works because the barrier uses a different `pSync` array, essentially an implicit double-buffering; we need to be careful here in case the implementation ever changes.

I think that using a full barrier between each isn't necessary, but fixing this more precisely requires some thought—I'll take a look after I'm done with #364, but this is fine as an interim fix.

I'm also interested in whether using the full barrier is better than looping over `while (atomic_fetch(ctx, &pSync[pe], pe) != ROCSHMEM_SYNC_VALUE)` prior to sending data to `pe`, or setting `pSync` with a compare-and-swap loop `while (atomic_compare_swap(ctx, &pSync[pe], ROCSHMEM_SYNC_VALUE, flag_val, pe) == ROCSHMEM_SYNC_VALUE)` to ensure that the remote PE is ready to receive data.

src/ipc/context_ipc_tmpl_device.hpp
omor1, line 393: Can there by sync issues between the two `internal_ring_allreduce`? Or between this `internal_ring_allreduce` and the `internal_direct_allreduce` below. They use the same `pSync` array, so I think they can run into the same ordering issue.

src/ipc/context_ipc_tmpl_device.hpp
edgargabriel: I have not tested this particular version, it might be worthwhile a test later. Its an interesting idea, but I am not sure how performant an atomic compare_and_swap to a remote memory location is.

src/ipc/context_ipc_tmpl_device.hpp
edgargabriel: that is a valid point. Let me think about that.

edgargabriel: actually, I forgot that I did look into this and resolved. This is the reason the wait_val is different in the first loop vs. the second loop.

omor1: The issue I was concerned about was between the two calls to `internal_ring_allreduce` within `reduce`, when `n_seg_up > n_seg`. If one PE is slow and resets `pSync` (at the end of the first ring-allreduce) _after_ another has written to that PE's `pSync`, then the slow PE will deadlock.

PR #386, opened 2026-01-14
Alltoallv

Yiltan:
Draft alltoallv implementation

AIROCSHMEM-218


src/gda/context_gda_tmpl_device.hpp
```diff
@@ -613,6 +613,52 @@ __device__ void GDAContext::alltoall(rocshmem_team_t team, T *dst,
...
+                                      const size_t source_displs[]) {
+  GDATeam *team_obj = reinterpret_cast<GDATeam *>(team);
+
+  int pe_start = team_obj->tinfo_wrt_world->pe_start;
+  int pe_size = team_obj->num_pes;
+  int stride = team_obj->tinfo_wrt_world->stride;
+  long *pSync = team_obj->alltoall_pSync;
+  int my_pe_in_team = team_obj->my_pe;
+  T *pWrk = reinterpret_cast<T *>(team_obj->pWrk);
+  size_t pWrk_offset = sizeof(double) * ROCSHMEM_REDUCE_MIN_WRKDATA_SIZE / sizeof(T);
```
nusislam: In an `alltoallv`, different ranks are sending different amount of data. How do you ensure that one rank is not overwriting the data from another rank in the staging buffer?

Yiltan: This is certainly a bug, I shall revisit this code/algorithm

src/gda/context_gda_tmpl_device.hpp
```diff
@@ -613,6 +613,52 @@ __device__ void GDAContext::alltoall(rocshmem_team_t team, T *dst,
...
+  for (int i = 0; i < pe_size; i++) {
+    int dest_pe = team_obj->get_pe_in_world(i);
+    T* src = source + source_displs[i];
+    T* dst = &pWrk[dest_pe * pWrk_offset];
+
+    if (source_nelems[i] == 0) {
+      continue;
+    }
+
+    put_wg(dst, src, source_nelems[i], dest_pe);
```
nusislam: Is this API accessible from RCCL?

Yiltan: Yes, it should map to `rocshmem_TYPENAME_put_wg`

https://rocm.docs.amd.com/projects/rocSHMEM/en/develop/api/rma.html#_CPPv424rocshmem_TYPENAME_put_wgP4TYPEPK4TYPE6size_ti

src/gda/context_gda_tmpl_device.hpp
```diff
@@ -613,6 +613,52 @@ __device__ void GDAContext::alltoall(rocshmem_team_t team, T *dst,
...
+  }
+
+  internal_sync_wg(my_pe, pe_start, stride, pe_size, pSync);
```
nusislam: What about this one? What equivalent API can RCCL use?

Yiltan: https://rocm.docs.amd.com/projects/rocSHMEM/en/develop/api/coll.html#_CPPv420rocshmem_sync_all_wgv

src/gda/context_gda_tmpl_device.hpp
```diff
@@ -613,6 +613,52 @@ __device__ void GDAContext::alltoall(rocshmem_team_t team, T *dst,
...
+                                      T *source, const size_t source_nelems[],
+                                      const size_t source_displs[]) {
+  GDATeam *team_obj = reinterpret_cast<GDATeam *>(team);
+
+  int pe_start = team_obj->tinfo_wrt_world->pe_start;
+  int pe_size = team_obj->num_pes;
+  int stride = team_obj->tinfo_wrt_world->stride;
+  long *pSync = team_obj->alltoall_pSync;
+  int my_pe_in_team = team_obj->my_pe;
+  T *pWrk = reinterpret_cast<T *>(team_obj->pWrk);
```
nusislam: How did you allocate `pWrk`? It should be a buffer registered across all PEs in the team.

Yiltan: pWrk is carved out of the symmetric heap. We do this since it needs to be registered so that we can issue a put/get operation to the remote pWrk buffer

Yiltan: to do this in RCCL you would do something like so:

```
void * pWrk = rocshmem_malloc(buff_size);
```

Then pass it into the kernel to use as a temp buffer

ammallya:
Imported to ROCm/rocm-systems pr 2730

PR #393, opened 2026-01-19
mlx5: remove `volatile' from store to dbrec

omor1:
Prior to #368 stores to dbrec were not volatile, restore this behavior.
See https://github.com/ROCm/rocSHMEM/pull/368#discussion_r2684437573 for details.


ammallya:
Imported to rocm/rocm-systems pr 2729

PR #395, opened 2026-01-21
[GDA/IONIC] only ring doorbell on active lanes

Yiltan:
## Motivation

Ringing doorbell was a bottleneck for performance

## Technical Details

Resolved by ringing doorbell only on active lanes

AIROCSHMEM-226

### DIFF
--- a/src/gda/ionic/queue_pair_ionic.cpp
+++ b/src/gda/ionic/queue_pair_ionic.cpp
@@ -204,8 +204,12 @@ __device__ void QueuePair::ionic_quiet_internal(uint64_t activemask, uint32_t co
 __device__ void QueuePair::ionic_ring_doorbell(uint32_t pos) {
   // When threads write at once to the same address, not all writes reach the bus.
   // Take turns and insert a thread fence between writes to the same address.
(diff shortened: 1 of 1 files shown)


omor1 (approved):
LGTM

I really dislike that the ionic code uses the term `activemask` to mean "the set of lanes that are using the same QP"—it's misleading, since the term usually means "the set of lanes that are set in the EXEC mask" and so it appears that e.g. `ionic_ring_doorbell` is called only by a single lane (see below); but this is an unrelated issue.
```HIP
__device__ uint32_t QueuePair::commit_sq(uint64_t activemask, uint32_t my_sq_prod, uint32_t my_sq_pos, uint32_t num_wqes) {
  uint32_t dbprod = my_sq_prod + num_wqes;

  spin_lock_acquire_shared(&sq_lock, activemask);

  if (is_first_active_lane(activemask) && ((sq_dbprod - dbprod) & (1u << 31))) {
    sq_dbprod = dbprod;

    ionic_ring_doorbell(dbprod);
  }

  spin_lock_release_shared(&sq_lock, activemask);

  return dbprod;
}
```

ammallya:
Imported to rocm/rocm-systems pr 2727

PR #399, opened 2026-01-25
add additional runtime checks and gfx1201 fix

edgargabriel:
add additional runtime checks and gfx1201 fix

For AIROCSHMEM-237

This commit contains three fixes:
 - increase the max. number of files at the beginning of the run to the max. allowed by the system
 - check for large BAR support. WE don not abort if its not available, but print a warning.
 - for gfx1201, do not use uncached memory at the moment.

(cherry picked from commit d21abd36efc1ef57ad2c391cca8fcae313fc359e)

fix allreduce tester (https://github.com/ROCm/rocSHMEM/pull/385)
- use the reduce_psync buffers for synchronization in allreduce, not the
  barrier_psync.
- execute a wwg barrier after the allreduce operation. After internal
  discussion it was determined that it is required for correctness.

(cherry picked from commit https://github.com/ROCm/rocSHMEM/commit/6f512e92a57719764c4ef5b071212c90e26c3ebe)

## Submission Checklist

### DIFF
--- a/src/bootstrap/bootstrap.cpp
+++ b/src/bootstrap/bootstrap.cpp
@@ -37,19 +37,6 @@
 
 namespace rocshmem {
 
-static void setFilesLimit() {
-  rlimit filesLimit;
-  if (getrlimit(RLIMIT_NOFILE, &filesLimit) != 0) {
-    DPRINTF("getrlimit failed\n");
-    return;
-  }
-  filesLimit.rlim_cur = filesLimit.rlim_max;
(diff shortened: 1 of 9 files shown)


edgargabriel:
This is supposed to be a cherry-pick, ideally no changes to be done compared to the original PR which was reviewed already. I will decline all Co-Pilot suggestions because of that.

abouteiller (dismissed):

abouteiller (approved):

